PR #21, opened 2017-03-28
Code signing

knuton:
This adds code signing steps to the package and release scripts. Code signing is done manually using a (hopefully short-lived) fork of [signcode](https://github.com/kevinsawicki/signcode) for two reasons:

1. `rcedit` bundled with electron-winstaller is broken and we already set the icon after the installer has been created, hence have to sign the installer ourselves too
2. Squirrel makes a hard-coded call to Microsoft's `signtool.exe`, which is bundled but does not run on Wine, so there is no out-of-the-box support for signing on non-Windows systems

Implementation is feature complete yet only tested with self-signed certificate.

#### Remaining

- [x] Test with real certificate

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -34,7 +34,7 @@ To start with Electron (as in production) run `npm run electron`.
 
 This will start the driver as an Electron application. Note that no window will be created. The application lives in the menubar.
 
-## Building Releases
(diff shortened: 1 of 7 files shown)


knuton:
Good to merge.

PR #24, opened 2017-04-19
standardjs

pukkamustard:
Reformat to standardjs

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -8,13 +8,9 @@ const path = require('path')
 
 // Run handler for Windows installer options
 if (require('./installation/handle-squirrel-events')()) {
-  // TODO: check if it makes sense to use process.exit() here. `return` causes a standardjs syntax error, as it can only be used in a function
   process.exit()
-  // return
 }
 
-// const Config = require('electron-config')
-// const config = new Config()
 const constants = require('./constants')
 
 const log = require('electron-log')
--- a/src/main.js
+++ b/src/main.js
@@ -1,73 +1,78 @@
-const electron = require('electron');
-const shell = electron.shell;
-const app = electron.app;
-const Menu = electron.Menu;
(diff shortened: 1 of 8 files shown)


pukkamustard:
I'm starting to really like standardjs...caught a few bugs just by reformating.

@knuton Can you check this in particular: https://github.com/dividat/driver/pull/24/files#diff-1c90ff38a08209f9ebd4d05d1e43358eR11

knuton:
Yes, I think the linter is right. The use of `return` was suggested in the electron-winstaller Readme or some such place, without further explanation. The fact that it works is due to an implementation detail of Node.

Using `process.exit` should be fine.

src/main.js
```diff
@@ -1,73 +1,78 @@
...
+  // TODO: check if it makes sense to use process.exit() here. `return` causes a standardjs syntax error, as it can only be used in a function
+  process.exit()
+  // return
+}
 
-const Config = require('electron-config');
-const config = new Config();
-const constants = require('./constants');
+// const Config = require('electron-config')
+// const config = new Config()
```
knuton: Why is this commented out?

src/main.js
```diff
@@ -1,73 +1,78 @@
...
+const Tray = electron.Tray
+
+const path = require('path')
 
 // Run handler for Windows installer options
-if (require('./installation/handle-squirrel-events')()) return;
+if (require('./installation/handle-squirrel-events')()) {
+  // TODO: check if it makes sense to use process.exit() here. `return` causes a standardjs syntax error, as it can only be used in a function
+  process.exit()
+  // return
```
knuton: Can be removed.

src/main.js
pukkamustard: Was commented out because unused in main.js. I removed it.

PR #26, opened 2017-05-05
"nearly always" Zeroconf

pukkamustard:
- Add error handling when mDNS can not bind or server can not listen
- Change default Senso address to dividat-senso.local

This makes setup "almost" or "nearly always" zeroconfig.

I chose not to add the bindings to system mDNS service (avahi, bonjour on windows/mac) and only use the javascript mDNS service. If a mDNS service is installed on the system, the JS mDNS client may fail. This is now handled and driver will not crash. In that case the default address of Senso has been set to "dividat-senso.local" and the system mDNS service will be able to resolve, driver will be able to connect.

If there is no system mDNS service, JS mDNS should work and Senso can be discovered and connected to.

For old installations that do not have mDNS on firmware yet, configurations are still read and basically nothing changes for them.

### DIFF
--- a/src/senso.js
+++ b/src/senso.js
@@ -5,7 +5,7 @@ const Connection = require('./persistentConnection')
 
 const DATA_PORT = 55568
 const CONTROL_PORT = 55567
-const DEFAULT_SENSO_ADDRESS = '169.254.1.10'
+const DEFAULT_SENSO_ADDRESS = 'dividat-senso.local'
 
 const log = require('electron-log')
 
@@ -40,7 +40,20 @@ var interfaces = R.filter(R.identity, R.flatten(R.values(R.map(R.map((ipInfo) =>
   }
 }), os.networkInterfaces()))))
 
-const bonjour = require('bonjour')({interface: interfaces})
+const bonjour = require('bonjour')({interface: interfaces, reuseAddr: true})
+
+// Handle case where binding to interface fails. NOTE: this is not standard multicast-dns API!
+bonjour._server.mdns.on('error', (error) => {
+  switch (error.code) {
+    case 'EADDRINUSE':
+      log.warn('MDNS: Could not bind to address ' + error.address)
+      break
+    default:
+      log.warn('MDNS: Error ' + error.code)
+      break
+  }
+})
+
 const bonjourOptions = {type: 'sensoControl'}
 
 function factory (sensoAddress, recorder) {
--- a/src/senso.js
+++ b/src/senso.js
@@ -27,34 +27,7 @@ try {
   }
 }
 
-// Set up MDNS client
-const os = require('os')
-const R = require('ramda')
-
-// compute list of interfaces to use for mdns client
-var interfaces = R.filter(R.identity, R.flatten(R.values(R.map(R.map((ipInfo) => {
-  if (ipInfo.internal === false && ipInfo.family === 'IPv4') {
-    return ipInfo.address
-  } else {
-    return false
-  }
-}), os.networkInterfaces()))))
-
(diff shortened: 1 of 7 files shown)


pukkamustard:
Situation where manual config would still be necessary:

JS mDNS fails and there is not system wide mDNS service installed.

pukkamustard:
> Can this TODO be removed?

Yes! done.

> This uses mDNS only partially, e.g. port numbers are still hardcoded. Does it make sense to change that? I guess we have two services, control and data.

The idea was to keep discovery of Senso just to IP address/host. No further information. But yes, I see that certain situations it would be nice to have that flexibility and do things _nicer_. However this also requires changes in Play: it needs to be capable of knowing that control and data are two different services provided by the Senso.

This is tricky. Let's discuss on Sunday.

> What happens when there are several Sensos on the local network? I think the driver connects to any one (maybe the first to respond?). We should offer a way of selecting Sensos instead, probably through Play.

Exactly. Driver auto-connects to first discovered Senso (https://github.com/dividat/driver/pull/26/files#diff-5713966e43ae05097c0f1086a0dcac1cR56).

Subsequent discovered Senos are notified to Play (https://github.com/dividat/driver/pull/26/files#diff-5713966e43ae05097c0f1086a0dcac1cR104). Play can already decode these messages (https://github.com/dividat/diviapps/pull/218) but doesn't do anything with them, yet.

I added comments to explain behavior more.

> Why connect to service.addresses[0] instead of service.host? With the latter, advertising a Senso from the replay script just works.

For reference what a `service` object looks like:
```
{ addresses: [ '10.42.0.249' ],
  name: 'Dividat Senso',
  fqdn: 'Dividat Senso._sensoControl._tcp.local',
  host: 'dividat-senso.local',
  referer: { address: '10.42.0.249', family: 'IPv4', port: 5353, size: 126 },
  port: 55567,
  type: 'sensoControl',
  protocol: 'tcp',
  subtypes: [],
  rawTxt: <Buffer >,
  txt: {} }
```
Using `service.host` would be the _correct_ way of doing things. However, it requires a system mDNS client to resolve `dividat-senso.local` to an IP (or additional code to resolve the dividat-senso.local hostname to an IP via mDNS in the driver). I chose to do the slightly less _correct_  but easier way.

pukkamustard:
> With the latter, advertising a Senso from the replay script just works.

Nice idea! I made it work like that (even without using `service.host`).

src/senso.js
```diff
@@ -40,7 +40,20 @@ var interfaces = R.filter(R.identity, R.flatten(R.values(R.map(R.map((ipInfo) =>
   }
 }), os.networkInterfaces()))))
 
-const bonjour = require('bonjour')({interface: interfaces})
+const bonjour = require('bonjour')({interface: interfaces, reuseAddr: true})
+
+// Handle case where binding to interface fails. NOTE: this is not standard multicast-dns API!
```
pukkamustard: This sets `SO_REUSEADDR` (or `SO_REUSEPORT`) which allows coexistance of system mDNS service and our mDNS client in certain situations (notably avahi on Linux).

On Mac coexistance with the system mDNS service does not work. I assume because the system service does not set `SO_REUSEPORT`:

> SO_REUSEPORT is what most people would expect SO_REUSEADDR to be. Basically, SO_REUSEPORT allows you to bind an arbitrary number of sockets to exactly the same source address and port as long as all prior bound sockets also had SO_REUSEPORT set before they were bound.
(http://stackoverflow.com/questions/14388706/socket-options-so-reuseaddr-and-so-reuseport-how-do-they-differ-do-they-mean-t)

knuton:
OK, great!

One more change request: Set host to `0.0.0.0` in `control.js` as well, to allow connecting through non-loopback IP.

PR #50, opened 2017-09-21
Unify platforms into common version

knuton:
This brings the functionality from `linux-link-local` back into the main line to prevent maintenance of separate branch lines.


pukkamustard:
Tested, works.

PR #56, opened 2017-11-17
Go release & update

etaque:
https://trello.com/c/p59Lv9g7/132-auto-update


pukkamustard:
I did a bit of reworking of the Makefile allowing stuff to be tested easier. It is now built up into 4 stages:
1. simple build: no cross compilation, this output is used for the test suite
2. cross compilation: cross compiles for all platforms but no signing or anything
3. release: builds the structure of the release in the release folder, signs the binaries
4. deploy: checks if version is tagged properly, if on correct branch and then deploys the build release to s3.

I don't like how many tests `check-version` was making and that it was impossible to build without an internet connection (for checking released version). Also the version in package.json is not relevant (as in the diviapps repository).

Release channel is set directly from branch name. The `deploy` step however only is allowed if you are on branch `master` or `develop` (https://github.com/dividat/driver/pull/56/files#diff-b67911656ef5d18c4ae36cb6741b7965R113).

The only additional check that I think is needed, is that version is tagged correctly. For that a simple regex in the makefile should do the trick. @knuton Some regex magic required. The unix tool for evalauating the regex is `expr`.

I also moved where the releases will live from releases/driver to releases/driver2 to avoid any conflict with the squirrel update mechanism for js driver that is still running.

Todos:
- [ ] Semver Regex
- [ ] Crosscompilation from Linux to Darwin does not "just work" (https://github.com/golang/go/issues/22462#issuecomment-339882044). Get it working.
- [ ] osslsign on linux

etaque:
> I don't like how many tests check-version was making and that it was impossible to build without an internet connection (for checking released version). 

But then you don't want to check if the released version is higher than the latest, before releasing?

etaque:
[Semver regex](https://github.com/sindresorhus/semver-regex/blob/master/index.js#L2)

PR #61, opened 2017-12-20
Run driver as service

pukkamustard:
Use https://github.com/kardianos/service to run dividat-driver as service.

Tested on Ubuntu with systemd.


pukkamustard:
[Driver as a service](https://trello.com/c/GMgpIwwT/135-driver-as-a-service)

PR #72, opened 2018-04-05
Statical linking againg musl

pukkamustard:
Unlike when compiling with "pure" go, cgo links dynamically (mostly to `glibc`, [1,2]). This usually works out ok, unless target host is using different version of `glibc` or it is in a completely different location (hey Nix!).

It is still possible to statically link with cgo, but it is a really bad idea to statically link `glibc` [3]. Enter [`musl`](https://www.musl-libc.org/) a standard library that was designed to be statically linked against.

These changes make builds on Linux statically link against musl. On Mac use `gcc` instead of Apple flavored `clang`.

For external C libraries (i.e. pcsc) this would require pcsc to be recompiled with musl.

[1] https://blog.hashbangbash.com/2014/04/linking-golang-statically/
[2] https://dominik.honnef.co/posts/2015/06/statically_compiled_go_programs__always__even_with_cgo__using_musl/
[3] https://tschottdorf.github.io/golang-static-linking-bug

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,18 @@ VERSION := $(shell git describe --always HEAD)
 # set the channel name to the branch name
 CHANNEL := $(shell git rev-parse --abbrev-ref HEAD)
 
-GO_LDFLAGS = -ldflags "-X dividat-driver/server.channel=$(CHANNEL) -X dividat-driver/server.version=$(VERSION) -X dividat-driver/update.releaseUrl=$(RELEASE_URL)"
+CC := gcc
+CXX := g++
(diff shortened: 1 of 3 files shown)


pukkamustard:
CI failes because of missing `musl-gcc`. I do not understand why it is not available on Travis...no signs of it being removed from `nixpkgs`.

pukkamustard:
Merging for now. WIll release 2.0.0-rc9 to test this out before releasing 2.0.0 next week.

@knuton Please note for pcsc.

PR #74, opened 2018-04-05
Set nix channel for Travis to 17.09

pukkamustard:
This fixes the nix channel Travis uses to 17.09 (stable).


pukkamustard:
woo-hoo! Travis happy, me happy!

PR #85, opened 2018-06-23
Build system

pukkamustard:
_Il nous faut de Nix, et encoure de Nix et tourjour de Nix._

Changes to build system:
- Lock version of nixpkgs
- Use nix to provide Go dependencies (new packages can be still installed with `dep` tooling is provided to sync `Gopkg.toml` with nix magic)
- Use the Nix cross-compilation machinery for building releases

This makes use of upstream changes in the nix cross compilation machinery. Conceptually releases are cross built for a very clearly specified system (e.g. `x86_64-linux-unknown-musl`) that may be different than the host system (e.g. `x86_64-linux-unknown-glibc` or `x86_64-darwin-*`). Nix takes care of providing the toolchain and dependencies for/in the target system.

Currently only one target system is defined (for Linux releases). Further target systems can be added.

I suggest running this as 2.1.0-RC2 for at least a week.

@knuton Can you test with card reader?


Makefile
```diff
@@ -164,29 +142,12 @@ deploy: release
...
-			--with-systemdsystemunitdir=$(PCSCLITE_DIR)/lib/systemd/system \
-			--disable-libudev \
-			--disable-libusb \
-			--disable-libsystemd; \
-		make; \
-		make install
-	rm -rf tmp/pcsclite
-	rm pcsc-1.8.23.tar.gz
-
+.PHONY: nix/deps.nix
```
knuton: Why phony?

Readme.md
```diff
@@ -29,37 +12,44 @@ Firefox, Safari and Edge not supported as they are not yet properly implementing
...
 
 To deploy a new release run: `make deploy`. This can only be done if you are on `master` or `develop` branch, have correctly tagged the revision and have AWS credentials set in your environment.
 
+### Go packages
+
+Go dependencies are provided by the [Go machinery](https://nixos.org/nixpkgs/manual/#sec-language-go) in Nix.
+
+For local development you may use `dep` to install go dependencies: `cd src/dividat-driver && dep ensure`.
+
+New Go dependencies can be added with `dep` (e.g. `dep ensure -add github.com/something/supercool`). The specification of dependencies needs to be recreated for Nix. This can be done with the command `make nix/deps.nix`. Check in the updated `Gopkg.toml`, `Gopkg.lock` and `nix/deps.nix` files.
```
knuton: Let's just make `crossbuild` (or rather the individual LINUX_BIN etc. targets) depend on `nix/deps.nix`?

knuton:
Good move for handling libpcsclite. Will check functionality with reader now.

Remarks so far:

- I think phoniness is a bit rampant in this Makefile, let's reduce that at least where it leaks into side remarks in the Readme.
- As we plan to still produce driver versions for all three platforms, also add definitions to build for Mac and Win.

Readme.md
```diff
@@ -29,37 +12,44 @@ Firefox, Safari and Edge not supported as they are not yet properly implementing
...
+
+### Cross compilation
+
+A default environment (defined in `default.nix`) provides all necessary dependencies for building on your native system (i.e. Linux or Darwin). Running `make` will create a binary that should run on your system (at least in the default environemnt).
```
knuton: This being under the header "cross compilation" is a bit confusing. Let's say very explicitly that crossbuilding is only supported on Linux, _but_ `make` will create a binary for the host system for development.

Makefile
pukkamustard: Good point. Added explicit dependency on `src/dividat-driver/Gopkg.toml`.

Readme.md
pukkamustard: I prefer having the default environment just being available. Usage of dep is not hindered in any way by providing already existing go packages via `nix/deps.nix`. Just makes it faster and easier to get started (on my machine `nix` is considerably faster than `dep ensure`).

Readme.md
pukkamustard: Yop, I merged the section "Cross compilation" with "Releasing" to subsections "Release/Building" and "Release/Deploying".

> I think phoniness is a bit rampant in this Makefile, let's reduce that at least where it leaks into side remarks in the Readme.

True. The root cause of this is that we do not really track root dependencies (i.e. go source files) with the Makefile but instead rely on the go compiler just being fast enough to compile everything (or do internal caching). It is a slightly perverted use of `Make` but I think it is legitimized in our case. Also by the fact that targets are mostly actions (e.g. `build`, `test`, `deploy`) which are really "phony" (i.e. not artifacts on the filesystem). Reducing phonyness requires restructuring the Makefile, which I don't think is absolutely necessary at this moment.

> As we plan to still produce driver versions for all three platforms, also add definitions to build for Mac and Win.

Makefile targets are there and in right format but commented out. I prefer adding them/uncommenting them when we have a working release target system definition.

pukkamustard:
sha256 issue should also be resolved now. I seem to have used a sha256 that was accepted by some versions of Nix (I use 2.0.2) but not others. Weird but resolved. Tested on Linux (Nix versions 2.0.2, 2.0 and1.11.15) and Darwin (Nix 1.11.16).

Readme.md
knuton: I'm okay with checking it in if there is a way to ensure that it stays up to date. I don't like these Readme side conditions on the build process that are easy to overlook or forget.

knuton: [Git hooks](https://git-scm.com/book/gr/v2/Customizing-Git-Git-Hooks) could be useful, but there doesn't seem to be a good way for sharing them.

knuton: And checked in or not: if we remove phoniness from the underlying targets, expressing the dependency on `nix/deps.nix` does not hurt execution time, but prevents confusion.

knuton:
PC/SC integration works fine 👍

knuton:
I continue the conversation here, as GitHub hides even new comments on outdated commits (hope this is a bug).

> Reducing phonyness requires restructuring the Makefile, which I don't think is absolutely necessary at this moment.

Reducing phoniness is worthwhile where having it implies more manual work or attention. Definitely worth removing it [here](https://github.com/dividat/driver/pull/85#discussion_r198070967). OK to leave `build` targets phony, `release` etc. obviously are.

> Makefile targets are there and in right format but commented out. I prefer adding them/uncommenting them when we have a working release target system definition.

I want to prevent platform drift. This changeset gives Nix a central place in the build system and motivates it with Nix's cross-compilation machinery. However it only provides one compilation target. I think the target system definition at least for macOS is pretty clear. Would be good to see the machinery in action.

Readme.md
pukkamustard: Very good point. And yup, it is very easily automatizable by adding `nix/deps.nix` as a dependency of the `build` target:

https://github.com/dividat/driver/pull/85/commits/537ed0d40a35aa3763e2d9dd43fd49a65f78d76d#diff-b67911656ef5d18c4ae36cb6741b7965L44

 Now on every build `nix/deps.nix` is regenerated if `src/dividat-driver/Gopkgs.toml` has been changed. Readme updated.

pukkamustard:
> Reducing phoniness is worthwhile where having it implies more manual work or attention. Definitely worth removing it here. OK to leave build targets phony, release etc. obviously are.

:ballot_box_with_check: 

> I want to prevent platform drift. This changeset gives Nix a central place in the build system and motivates it with Nix's cross-compilation machinery. However it only provides one compilation target. I think the target system definition at least for macOS is pretty clear. Would be good to see the machinery in action.

Absolutely. This changeset is very much in preparation of build targets for Darwin/Windows. They will follow in subsequent changesets.

From my side I think this is ready to merge. @knuton Please confirm.

knuton:
Good, confirmed.

PR #86, opened 2018-07-04
Fix string representation of RFID UIDs

knuton:
Small byte values were not zero-padded when printed individually.

Instead, the whole byte array can be printed in hexadecimal.


src/dividat-driver/rfid/pcsc.go
```diff
@@ -197,8 +197,8 @@ func waitForCardActivity(haveBeenKilled *bool, log *logrus.Entry, scard_ctx *sca
 				continue
 			}
 			uid := ""
-			for i := 0; i < len(response)-2; i++ {
-				uid += fmt.Sprintf("%X", response[i])
+			if len(response) > 2 {
```
knuton: Use constant `ERROR_FLAG_LENGTH = 2` o.s. to explain dropping last two values.

PR #89, opened 2018-07-17
[TRAVIS-QUESTION] Hi Travis! Can you cache the nix-store?

pukkamustard:


pukkamustard:
Travis does not use cache :cry: ... but it might have something to do with being on the right branch (or this branch is not considered important enough to create a new cache entry (see: https://docs.travis-ci.com/user/caching/#Pull-request-builds-and-caches).

PR #97, opened 2018-09-19
Add a target to build without dep2nix, for homebrew.

etaque:
Keeps the nix deps as default target, extracts the simple build as a target for homebrew build, for which we can't require dep2nix and who's randomly confused by nix/deps.nix timestamp at checkout.


pukkamustard:
I think better is to simply remove the `nix/deps.nix` dependecy from the `build` target. It does not make so much sense there as recreating the `nix/deps.nix` file will not change the already loaded environment and might cause the loaded (and checked in) environment to differ from what is specified in `nix/deps.nix`.

The idea of adding this was to automate the recreation of the `deps.nix` file. But I think a better place to do it would be as a git-hook or as a reminder when creating a pull request.

etaque:
Ok, pushed the removal of it. I guess we'll have to update it during dev loop anyway, otherwise we wouldn't see deps changes.

PR #107, opened 2021-02-24
Add Senso Flex bridge

knuton:
This adds a new WebSocket bridge for Flex devices. The basic idea is the same as for Senso and RFID, we close the hardware gap for the web application, in this case relaying serial device input via WebSocket messages.

The branch includes commits that capture the developmental stages of the Flex prototypes:

- 18766d6: First 4x4 developer kit (finger-sized)
- 8aa4b0b: 16x16 sensor developer kit
- 82f099f: First custom built mats (push-based protocol)
- 589dc9a: Second and up to current custom built mats (pull-based protocol)

Further iterations on the device protocol are expected, but I think this is stable and robust enough to be integrated.

### Testing

Unfortunately there isn't much to test without a Flex device. I tried to provide a replayable mock device using [umockdev](https://github.com/martinpitt/umockdev), but older versions I tried didn't capture anything, while versions 0.13.0 and up segfault when the tty device is accessed. I'd like to modernize the driver package to use a current version of Go and it might be worth trying again then.

### DIFF
--- a/src/dividat-driver/flex/main.go
+++ b/src/dividat-driver/flex/main.go
@@ -182,8 +182,8 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
 
 	reader := bufio.NewReader(port)
 	state := WAITING_FOR_HEADER
-	samplesLeftInSet := 0
-	bytesLeftInSample := 0
+	var samplesLeftInSet int
+	var bytesLeftInSample int
 
 	var buff []byte
 	for {
@@ -217,6 +217,7 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
 			state = BODY_START
 		case state == BODY_START && input == '\n':
 			state = BODY_READ_SAMPLE
+			buff = []byte{}
 			bytesLeftInSample = BYTES_PER_SAMPLE
 		case state == BODY_READ_SAMPLE:
 			buff = append(buff, input)
@@ -230,7 +231,6 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
 					onReceive(buff)
 
 					// Get ready for next set and request it
-					buff = []byte{}
 					state = WAITING_FOR_HEADER
 					_, err = port.Write(START_MEASUREMENT_CMD)
 					if err != nil {
@@ -244,9 +244,6 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
 			}
 		case state == UNEXPECTED_BYTE && input == HEADER_START_MARKER:
 			// Recover from error state when a new header is seen
-			buff = []byte{}
-			bytesLeftInSample = 0
-			samplesLeftInSet = 0
 			state = HEADER_START
 		default:
 			state = UNEXPECTED_BYTE
--- a/src/dividat-driver/flex/main.go
+++ b/src/dividat-driver/flex/main.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"sync"
 
+	"encoding/binary"
 	"bufio"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"path"
@@ -130,18 +130,18 @@ func scanAndConnectSerial(ctx context.Context, logger *logrus.Entry, onReceive f
 type ReaderState int
 
 const (
-	// Init
-	WaitingForFirstHeader ReaderState = iota
-	// Header
-	HeaderStarted
-	ExpectingHeaderEnd
-	// Data row
-	RowStarted
-	WaitingForRowIndex
-	ReadingRowData
-	ReachedRowEnd
-	// Error state
-	UnexpectedByte
+	WAITING_FOR_HEADER ReaderState = iota
+	HEADER_START
+	HEADER_READ_LENGTH_MSB
+	WAITING_FOR_BODY
+	BODY_START
+	BODY_POINT
+	UNEXPECTED_BYTE
+)
+
+const (
+	HEADER_START_MARKER = 'N'
+	BODY_START_MARKER = 'P'
 )
 
 // Actually attempt to connect to an individual serial port and pipe its signal into the callback, summarizing
@@ -173,8 +173,9 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
 	}
 
 	reader := bufio.NewReader(port)
-	state := WaitingForFirstHeader
-	bytesLeftInRow := 0
+	state := WAITING_FOR_HEADER
+	pointsLeftInSet := 0
+	bytesLeftInPoint := 0
 
 	var buff []byte
 	for {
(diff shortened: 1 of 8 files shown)


guyonvarch:
Well done!

I had a look at each commit in the history, and started looking at the final diff. I begin to have a good understanding on the changeset.

I’ll re-look at it tomorrow.

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+// Actually attempt to connect to an individual serial port and pipe its signal into the callback, summarizing
+// package units into a buffer.
+func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string, onReceive func([]byte)) {
+	mode := &serial.Mode{
+		BaudRate: 115200,
+		Parity:   serial.NoParity,
+		DataBits: 8,
+		StopBits: serial.OneStopBit,
```
guyonvarch, line 162: I am not accustomed to serial ports, how did you pick those values ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+	if err != nil {
+		logger.WithField("config", mode).WithField("error", err).Info("Failed to open connection to serial port.")
+		return
+	}
+	defer port.Close()
+
+	_, err = port.Write(START_MEASUREMENT_CMD)
+	if err != nil {
+		logger.WithField("error", err).Info("Failed to write start message to serial port.")
+		port.Close()
```
guyonvarch, line 178: This should already be called thanks to the _defer_ above ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+		return
+	}
+
+	reader := bufio.NewReader(port)
+	state := WAITING_FOR_HEADER
+	pointsLeftInSet := 0
+	bytesLeftInPoint := 0
```
guyonvarch, line 185: What are points and sets referred in these namings ? Are points individual captors in the flex mat ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+
+	var buff []byte
+	for {
+		// Terminate if we were cancelled
+		if ctx.Err() != nil {
+			logger.WithField("name", serialName).Info("Disconnecting from serial port.")
```
guyonvarch, line 191: Should this log be moved along `port.Close` in the _defer_ ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+
+
+// Serial communication
+
+type ReaderState int
+
+const (
+	WAITING_FOR_HEADER ReaderState = iota
+	HEADER_START
+	HEADER_READ_LENGTH_MSB
```
guyonvarch, line 143: What is _MSB_ ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+		switch {
+		case state == WAITING_FOR_HEADER && input == HEADER_START_MARKER:
+			state = HEADER_START
+		case state == HEADER_START && input == '\n':
+			state = HEADER_READ_LENGTH_MSB
+                case state == HEADER_READ_LENGTH_MSB:
+			// The number of measurements in each set may vary and is
+			// given as two consecutive bytes (big-endian).
+			msb := input
+			lsb, err := reader.ReadByte()
```
guyonvarch, line 212: What does _lsb_ mean ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+					// Finish and send set
+					onReceive(buff)
+
+					// Get ready for next set and request it
+					buff = []byte{}
+					state = WAITING_FOR_HEADER
+					_, err = port.Write(START_MEASUREMENT_CMD)
+					if err != nil {
+						logger.WithField("error", err).Info("Failed to write poll message to serial port.")
+						port.Close()
```
guyonvarch, line 242: Also not needed ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+			} else if err != nil {
+				continue
+			}
+			pointsLeftInSet = int(binary.BigEndian.Uint16([]byte{msb,lsb}))
+			state = WAITING_FOR_BODY
+		case state == WAITING_FOR_BODY && input == BODY_START_MARKER:
+			state = BODY_START
+		case state == BODY_START && input == '\n':
+			state = BODY_POINT
+			bytesLeftInPoint = 4
```
guyonvarch, line 224: Use a constant for 4 ? It could also be used below, when starting the next point and reinitializing the variable.

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+				} else {
+					// Start next point
+					bytesLeftInPoint = 4
+				}
+			}
+		case state == UNEXPECTED_BYTE && input == HEADER_START_MARKER:
+			// Recover from error state when a new header is seen
+			buff = []byte{}
+			bytesLeftInPoint = 0
+			pointsLeftInSet = 0
```
guyonvarch, line 254: It should be set to the right values afterward ?

src/dividat-driver/flex/main.go
guyonvarch: Also reset `buff` here ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+			state = HEADER_START
+		default:
+			state = UNEXPECTED_BYTE
```
guyonvarch, line 257: Debug log unexpected bytes ? Or would it be too chatty ?

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		input, err := reader.ReadByte()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			continue
+		}
+
+		// Finite State Machine for parsing byte stream
```
guyonvarch, line 202: Is there a specification somewhere telling how it works ?

guyonvarch (changes requested):

src/dividat-driver/flex/main.go
```diff
@@ -0,0 +1,262 @@
...
+// package units into a buffer.
+func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string, onReceive func([]byte)) {
+	mode := &serial.Mode{
+		BaudRate: 115200,
+		Parity:   serial.NoParity,
+		DataBits: 8,
+		StopBits: serial.OneStopBit,
+	}
+
+	START_MEASUREMENT_CMD := []byte{'S', '\n'}
```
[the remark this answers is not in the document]
knuton: Unfortunately not possible in Go with a byte array, agree that it would be nice.

src/dividat-driver/flex/main.go
knuton: Most Significant Byte

src/dividat-driver/flex/main.go
knuton: Least Significant Byte

src/dividat-driver/flex/main.go
knuton: Yes, the idea is to reset to noop values and rely on later stages setting the appropriate values (some of which are read from serial data).

src/dividat-driver/flex/main.go
knuton: I don't see the need at this moment. If there was a serious issue there, we would probably further instrument the driver to observe.

src/dividat-driver/flex/main.go
knuton: Kind of. Uploaded it to FW repo.

src/dividat-driver/flex/main.go
knuton: Renamed "point" to "sample": 476cd27

"Set" might then be clear enough? The device sends sets of samples from the sensors, measured at ~ the same moment in time.

src/dividat-driver/flex/main.go
knuton: Thanks, I unified these according to your suggestions: 24c1abd

src/dividat-driver/flex/main.go
knuton: Good idea, introduced a constant in c7902cc.

Not sure about resetting `buff`. Just for good measure? I think this shouldn't be needed unless there are mistaken writes to it elsewhere.

knuton:
Testing the `defer` changes I found and fixed a potential bug: a8599a9

Back to you. :slightly_smiling_face:

src/dividat-driver/flex/main.go
guyonvarch: :+1:

guyonvarch:
It’s a good thing you caught a bug while testing!

All good, last thing remaining is state initialization, a commit should be clearer than the 2 messages I wrote (non resolved messages): https://github.com/guyonvarch/driver/commit/4bf5abc04337e4d70406aad454f7da421d75e4a1

PR #108, opened 2021-04-14
Use Go modules instead of dep

guyonvarch:
dep has been deprecated:

> NOTE: Dep was an official experiment to implement a package manager
> for Go. As of 2020, Dep is deprecated and archived in favor of Go
> modules, which have had official support since Go 1.11. For more
> details, see https://golang.org/ref/mod.

https://github.com/golang/dep

The reason I am doing the migration now is I have issues with `dep` to
download the `go-bluetooth` dependency. Searching in `dep` issues, the
error I get look the same as this one:

`https://github.com/golang/dep/issues/2240`

The proposed resolution is to migrate to Go modules.

## Tests done

- `make` and `make crossbuild` are producing binaries
- `Senso flex` signal is transmitted as expected
- `./build vm` is working in PlayOS with the following derivation:

```nix
{stdenv, fetchFromGitHub, buildGoModule, pkgs}:

buildGoModule rec {
  pname = "dividat-driver";
  version = "2.1.0";

  src = /path/to/driver;

  vendorSha256 = "1lvgp9q3g3mpmj6khbg6f1z9zgdlmwgf65rqx4d7v50a1m7g9a0m";

  nativeBuildInputs = with pkgs; [ pkgconfig pcsclite ];
  buildInputs = with pkgs; [ pcsclite ];
}
```

## Version upgrades

I did not force version in `go.mod`, it may have taken the last available ones. We can try to use older versions, but it may be also an opportunity to upgrade some versions.

## Remaining work

I am keeping this PR as a draft, there may be more things to do. For example, `default.nix` depends on `deps.nix`, which does not exist anymore.


knuton:
:+1: If it's useful, let's switch.

> I did not force version in `go.mod`, it may have taken the last available ones. We can try to use older versions, but it may be also an opportunity to upgrade some versions.

Yes, let's upgrade. Appropriate system testing will be in order.

> I am keeping this PR as a draft, there may be more things to do. For example, `default.nix` depends on `deps.nix`, which does not exist anymore.

So `default.nix` could be rewritten to a `shell.nix` providing appropriate binaries and we could stop using `buildGoPackage`, which we do not properly use in this project at the moment.

guyonvarch:
> So default.nix could be rewritten to a shell.nix providing appropriate binaries and we could stop using buildGoPackage, which we do not properly use in this project at the moment. PlayOS would use a separate definition as you outlined above.

https://github.com/dividat/driver/pull/108/commits/625e397edea1f9315c49d0069bf7399402f0fd75

I had a try upgrading `nixpkgs` to `20.09` and to `unstable` of today, but I am getting an error with `make crossbuild` each time. So, keeping the old `nixpkgs` for now.

nix/nixpkgs.nix
```diff
@@ -1,94 +1,87 @@
-let
```
guyonvarch, line 1: Hiding whitespace changes simplifies the diff in this file.

knuton (changes requested):
The version and channel variables seem to not be set by the linker anymore:

```
> curl localhost:8382
{"arch":"amd64","channel":"","machineId":"<redacted>","message":"Dividat Driver","os":"linux","version":""}
```

This should happen here: https://github.com/guyonvarch/driver/blob/c518e05fb563ea6fef34f5d708a11cd55c0d9ac4/Makefile#L34

https://github.com/golang/go/wiki/GcToolchainTricks#including-build-information-in-the-executable

guyonvarch:
> The version and channel variables seem to not be set by the linker anymore:

Exact, this is fixed with https://github.com/dividat/driver/pull/108/commits/6019bea3b3b330042b4df8d53b0b98a322bf3fe5:

```json
{
  "arch": "amd64",
  "channel": "go-modules",
  "machineId": "<machineId>",
  "message": "Dividat Driver",
  "os": "linux",
  "version": "2.2.0-rc1-36-gc518e05"
}
```

The paths are a bit longer now.

knuton (approved):
Because of possible depedency bump I did a basic functionality check for RFID and Senso discovery under Windows, Senso discovery under Linux.

PR #109, opened 2021-04-14
Add PR template

guyonvarch:
I noticed that there is a `Changelog.md` file. We may have missed to add entries to it lately, in particular the recent `Senso flex` addition ?


knuton (approved):
:+1:

PR #110, opened 2021-04-26
Connect to Senso flex via bluetooth

guyonvarch:
Monitor bluetooth devices in a daemon, in order to keep a list of connected devices.

After having scanned serial ports, try to connect via Bluetooth to a Senso Flex device.

## Testing

1. Enable Bluetooth in OS, in nixos: `hardware.bluetooth.enable = true;`.
2. Activate bluetooth in network settings.
3. Build and run the driver.
4. Run Play at `dev-play.dividat.com`.
5. Power on Senso Flex without using the USB link, the led should blink.
6. After some seconds, the led should be fixed, this means Senso Flex is connected.

## Improvements

- Powering off Senso Flex, and then powering on again, it takes a long time to use it again, because the socket timeout is long.
- The flush function in some places, this works but it may not be the best solution.
- Logging in bluetooth monitoring may follow what’s used in other places in the app.

## Checklist

-   [ ] Changelog updated
-   [x] Code documented


knuton:
Great, could get it to connect easily!

The connection seems a bit sluggish to me, however, I will have to test alternative room arrangements and also try to quantify the sluggishness.

guyonvarch:
> I experience noticeable variances in signal quality while using the mat via Bluetooth. This occurs with mat and laptop close to each other and without insulating materials separating them. Sometimes there is no noticeable lag, sometimes it is visible with the bare eye. To crosscheck that it isn't due to some changes in the driver implementation, I confirmed that throughput on a wired connection continues to be good in this branch.

I think I also notice it. It’s not always very visible, but I see the difference looking at the vizualisation. I also played Hexagons:

- with bluetooth, it seems the movement is done just after the step is done,
- whereas with a cable, it seems almost immediate.

knuton:
Closing this as further work on the mat device will be necessary before Bluetooth can be explored in more depth.

PR #112, opened 2021-05-27
Use Nix mirror for sourceforge archive

knuton:


knuton:
A simpler and more durable solution is to fastlane code signing removal for Windows: https://github.com/dividat/driver/pull/113

PR #114, opened 2021-05-27
Remove self update

knuton:
A quick go at ripping out self-update functionality.

https://github.com/dividat/driver/commit/fe8dc03cd3b5cdbc58f5edfdce4337e9f664509b would be a straightforward way of being able to still release binaries for convenience of external collaborators, but without impacting any legacy installations.

We could also consider to stop publishing driver binaries to `dist.dividat.com` and just attach binaries to releases here on GitHub, which we could have built through GitHub Actions. Benefit: Easy browsing of binaries and sources together. Downside: More long-term dependence on GitHub.

## Checklist

-   [x] Changelog updated
-   [x] Code documented


Makefile
```diff
@@ -20,8 +20,8 @@ OUT ?= bin/dividat-driver
 # Get version from git
 VERSION := $(shell git describe --always HEAD)
 
-# set the channel name to the branch name
-CHANNEL := $(shell git rev-parse --abbrev-ref HEAD)
+# Only channel is main now
+CHANNEL := main
```
guyonvarch, line 24: The information can be updated as well with `make deploy`:

```
	# Print information about channel heads and confirm
	@echo "Channel 'master' is at:"
	@echo "  $(shell git show --oneline --decorate --quiet $$(curl -s "$(RELEASE_URL)master/latest" | tr -d '\n') | tail -1)"
	@echo "Channel 'develop' is at:"
	@echo "  $(shell git show --oneline --decorate --quiet $$(curl -s "$(RELEASE_URL)develop/latest" | tr -d '\n') | tail -1)"
	@echo
	@echo "About to deploy $(VERSION) to '$(CHANNEL)'. Proceed? [y/N]" && read ans && [ $${ans:-N} == y ]
```

Makefile
```diff
@@ -33,8 +33,7 @@ endif
 
 GO_LDFLAGS = -ldflags "$(STATIC_LINKING_LDFLAGS) \
 						 -X github.com/dividat/driver/src/dividat-driver/server.channel=$(CHANNEL) \
```
guyonvarch, line 35: Because we now only have main, we could remove this information that seems to be used only in the logs.

guyonvarch (commented):
The driver is still working as expected, tested with the relay and also running the tests.

Makefile
knuton: Thanks, I pushed an updated 47fcc67e99f7fe2c8c40c80bc98029ff3d0994a4.

Makefile
knuton: Fair enough: fe6405b

guyonvarch (approved):
LGTM

PR #124, opened 2023-07-06
Make firmware update command optimistic

knuton:
Instead of giving up immediately when we can not detect a Senso in
regular mode or can not send the DFU command, keep going and try to
discover a Senso in bootloader mode that we may be able to transmit the
firmware image to.

This helps us recover from a scenario where a Senso ended up in
bootloader mode but did not receive the firmware due to some network
issue.

Sensos up to at least firmware 3.8.1 do not restart into the regular
mode once brought to bootloader mode, so leaving them there without
means for recovery is problematic.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -75,6 +75,9 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 	if controllerHost != "" {
 		err := sendDfuCommand(controllerHost, controllerPort)
 		if err != nil {
+			// Log the failure, but continue anyway, as the Senso might have been left in
+			// bootloader mode when a previous update process failed. Not all versions of
+			// the firmware automtically exit from the bootloader mode upon restart.
 			fmt.Printf("Could not send DFU command to Senso at %s: %s\n", controllerHost, err)
 		}
 	} else {
@@ -90,7 +93,12 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		discoveredAddr, err := discover("_sensoUpdate._udp", deviceSerial, ctx)
 		cancel()
 		if err != nil {
-			// Try to discover boot controller via legacy identifier
+			// Up to firmware 2.0.0.0 the bootloader advertised itself with the same
+			// service identifier as the application level firmware. To support such
+			// legacy devices, we look for `_sensoControl` again at this point, if
+			// the other service has not been found.
+			// We do need to rediscover, as the legacy device may still just have
+			// restarted into the bootloader and obtained a new IP address.
 			ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
 			legacyDiscoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
 			cancel()
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -45,6 +45,8 @@ func Command(flags []string) {
 	err = Update(context.Background(), file, deviceSerial, configuredAddr)
 	if err != nil {
 		fmt.Println(err.Error())
+		fmt.Println()
+		fmt.Println("Update failed. Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again.")
 		os.Exit(1)
 	}
 }
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -108,7 +108,7 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		fail = err
 		return
 	}
-	fmt.Println("✓ Firmware transmitted to Senso.")
+	fmt.Println("Success! Firmware transmitted to Senso.")
 	return
 }
 
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
-	"errors"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


src/dividat-driver/firmware/main.go
```diff
@@ -86,29 +94,29 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 			ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
 			legacyDiscoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
```
krksgbr, line 95: Indeed as @guyonvarch suggested, trying to re-discover here with `_sensoControl._tcp` might not make sense. At this point we've already tried that in step one. I think it doesn't hurt, however unless a Senso could somehow switch back from DFU mode to normal mode between step 1 and 3 - which I don't think is the case - it seems unnecessary.

krksgbr: Maybe something like this would make it more obvious?

```diff
diff --git a/src/dividat-driver/firmware/main.go b/src/dividat-driver/firmware/main.go
index aad158b..6b80743 100644
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -62,7 +62,7 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 	} else {
 		// Discover controller address via mDNS
 		ctx, cancel := context.WithTimeout(parentCtx, 15*time.Second)
-		discoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
+		discoveredAddr, err := discoverSensoInNormalMode(deviceSerial, ctx)
 		cancel()
 		if err != nil {
 			fmt.Printf("Error: %s\n", err)
@@ -87,12 +87,12 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		dfuHost = *configuredAddr
 	} else {
 		ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
-		discoveredAddr, err := discover("_sensoUpdate._udp", deviceSerial, ctx)
+		discoveredAddr, err := discoverSensoInDFUMode(deviceSerial, ctx)
 		cancel()
 		if err != nil {
 			// Try to discover boot controller via legacy identifier
 			ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
-			legacyDiscoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
+			legacyDiscoveredAddr, err := discoverSensoInNormalMode(deviceSerial, ctx)
 			cancel()
 			if err == nil {
 				dfuHost = legacyDiscoveredAddr
@@ -172,6 +172,15 @@ func putTFTP(host string, port string, image io.Reader) error {
 	return nil
 }
 
+
+func discoverSensoInDFUMode(deviceSerial *string, ctx context.Context)(addr string, err error){
+  return discover("_sensoUpdate._udp", deviceSerial, ctx)
+}
+
+func discoverSensoInNormalMode(deviceSerial *string, ctx context.Context)(addr string, err error){
+  return discover("_sensoControl._tcp", deviceSerial, ctx)
+}
+
 func discover(service string, deviceSerial *string, ctx context.Context) (addr string, err error) {
 	resolver, err := zeroconf.NewResolver(nil)
 	if err != nil {
```

src/dividat-driver/firmware/main.go
```diff
@@ -86,29 +94,29 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
...
+	// 4: Transmit firmware via TFTP
+	time.Sleep(5 * time.Second) // Wait to ensure proper TFTP startup
+	err := putTFTP(dfuHost, tftpPort, image)
 	if err != nil {
 		fail = err
 		return
 	}
-	fmt.Println("✓ Firmware transmitted to Senso.")
+
+	fmt.Println("Success! Firmware transmitted to Senso.")
```
krksgbr, line 119: At this point we could try to transmit to a Senso to which we failed to send a DFU command.
Adding some print statements and using a non-sensical address. I added some additional print statements to better see what the code is doing: 
```diff
diff --git a/src/dividat-driver/firmware/main.go b/src/dividat-driver/firmware/main.go
index aad158b..8ef3503 100644
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -61,6 +61,7 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		fmt.Printf("Using specified controller address '%s'.\n", controllerHost)
 	} else {
 		// Discover controller address via mDNS
+		fmt.Printf("No controller address specified. Trying to find Senso via mDNS.\n")
 		ctx, cancel := context.WithTimeout(parentCtx, 15*time.Second)
 		discoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
 		cancel()
@@ -68,11 +69,13 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 			fmt.Printf("Error: %s\n", err)
 		} else {
 			controllerHost = discoveredAddr
+		  fmt.Printf("Discovered Senso at: %s\n", controllerHost)
 		}
 	}
 
 	// 2: Switch the Senso to bootloader mode
 	if controllerHost != "" {
+    fmt.Printf("Sending DFU command to Senso at: %s\n", controllerHost)
 		err := sendDfuCommand(controllerHost, controllerPort)
 		if err != nil {
 			fmt.Printf("Could not send DFU command to Senso at %s: %s\n", controllerHost, err)
@@ -87,11 +90,14 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		dfuHost = *configuredAddr
 	} else {
 		ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
+    fmt.Printf("Trying to find Senso in DFU mode.\n", controllerHost)
 		discoveredAddr, err := discover("_sensoUpdate._udp", deviceSerial, ctx)
 		cancel()
 		if err != nil {
 			// Try to discover boot controller via legacy identifier
 			ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
+      fmt.Printf("Could not find Senso in DFU mode.\n", controllerHost)
+      fmt.Printf("Trying to find Senso via legacy identifier.\n", controllerHost)
 			legacyDiscoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
 			cancel()
 			if err == nil {
@@ -105,13 +111,18 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 			}
 		} else {
 			dfuHost = discoveredAddr
+      fmt.Printf("Discovered Senso in DFU mode at: %s\n", dfuHost)
 		}
 	}
 
+  fmt.Printf("Waiting 5 seconds for TFTP startup.\n")
 	// 4: Transmit firmware via TFTP
 	time.Sleep(5 * time.Second) // Wait to ensure proper TFTP startup
+
+  fmt.Printf("Transferring image to Senso at: %s.\n", dfuHost)
 	err := putTFTP(dfuHost, tftpPort, image)
 	if err != nil {
+    fmt.Printf("Failed to transfer image.\n")
 		fail = err
 		return
 	}
```

Then when I update using:
```
./bin/dividat-driver update-firmware -i <image> -a "nonaddress"
```

I see:
```
Using specified controller address 'nonaddress'.
Sending DFU command to Senso at: nonaddress
Could not send DFU command to Senso at nonaddress: Could not dial connection to Senso controller at nonaddress:55567: dial tcp: lookup nonaddress: no such host
Waiting 5 seconds for TFTP startup.
Transferring image to Senso at: nonaddress.
Failed to transfer image.
Could not create tftp client: resolving address nonaddress:69: lookup nonaddress: no such host

Update failed. Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again.
```

src/dividat-driver/firmware/main.go
knuton: Oh, I see. In firmware versions before 2.0.0.0 the bootloader advertised itself as `_sensoControl` instead of `_sensoUpdate`.

This is what is meant by "legacy" here, but we should add a more explicit comment.

src/dividat-driver/firmware/main.go
knuton: > At this point we could try to transmit to a Senso to which we failed to send a DFU command.

Yes, that is part of the cockroach strategy. Try everything in order to recover and pick up from any point at which we previously might have gotten stuck.

In this case, if a Senso was sent to update/bootloader mode before, it is going to remain there (it seems even past a reboot), so it _is_ going to be in bootloader mode and wait for the TFTP connection, but it is not going to offer the connection to receive the DFU command.

knuton:
I think I should add more documentation about the cockroach strategy and specific assumptions first.

krksgbr (approved):
Agree, the code could benefit from refactoring but it gets the job done.
Given the time constraints, let's make progress with the rest of the project and come back to refactor when we're on a good track.

PR #127, opened 2024-03-28
DIV-33 Update firmware via websocket command

krksgbr:
Changes at a high level:
The messaging protocol was extended to allow receiving a command initiating a firmware update. The command carries a base64 encoded binary of the firmware which is decoded and sent to the Senso. During the update process the driver sends progress messages back to the client.

The discovery process was re-factored to allow simultaneous discovery of Sensos that could be in both "Application" and "Bootloader" mode. The discovery process now also checks the `mode` txt entry to determine what state a Senso could be in, as older versions of the firmware announce themselves as `_sensoControl._tcp` instead of `_sensoUpdate._udp` when in bootloader mode.

Checking the authenticity of the firmware image is still to be added.

## Checklist

-   [X] Changelog updated
-   [X] Code documented


src/dividat-driver/senso/websocket.go
```diff
@@ -293,6 +338,12 @@ func (handle *Handle) dispatchCommand(ctx context.Context, log *logrus.Entry, co
 
 		return nil
 
+	} else if command.UpdateFirmware != nil {
+		handle.ProcessFirmwareUpdateRequest(*command.UpdateFirmware, func(msg FirmwareUpdateMessage) {
+			sendMessage(Message{
+				FirmwareUpdateMessage: &msg,
+			})
+		})
```
krksgbr, line 346: I think we need to track if an update is in progress and ignore commands that come in the meantime.
To do that, we need to process the update command asynchronously (right now all incoming messages are processed synchronously) so we can keep listening to incoming websocket messages and actively ignore them.

src/dividat-driver/senso/update_firmware.go
```diff
@@ -0,0 +1,48 @@
...
+	"bytes"
+	"context"
+	"encoding/base64"
+	"fmt"
+	"io"
+
+	"github.com/dividat/driver/src/dividat-driver/firmware"
+)
+
+type OnUpdate func(msg FirmwareUpdateMessage)
```
krksgbr, line 13: I used a channel instead of a callback previously, but ran into issues with some messages not being sent and others being sent twice. I couldn't figure out why that happened so I simplified to use a callback instead.

src/dividat-driver/senso/websocket.go
krksgbr: [Ignore commands while firmware update is in progress](https://github.com/dividat/driver/pull/127/commits/97fd00550e6f31bd8dddbca12c2dbddc9b6b68b7)

src/dividat-driver/firmware/main.go
```diff
@@ -90,7 +97,7 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		dfuHost = *configuredAddr
 	} else {
 		ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
-		discoveredAddr, err := discover("_sensoUpdate._udp", deviceSerial, ctx)
+		discoveredAddr, err := discover("_sensoUpdate._udp", deviceSerial, ctx, onProgress)
```
krksgbr, line 100: The address could change here, right? Should we compare serial numbers to see if we're rediscovering the same Senso as the one we meant to update?

src/dividat-driver/senso/update_firmware.go
knuton: The callback looks good to me :+1:

src/dividat-driver/firmware/main.go
knuton: I think it would be best to start the update process by a device serial passed from the client to the driver, as the address indeed might change.

It would be good to also allow passing a fixed address, though, which Play could do if a connection is configured in this way instead of via serial. (I think it would be OK to only offer the serial option initially.)

src/dividat-driver/senso/update_firmware.go
```diff
@@ -0,0 +1,66 @@
...
+		msg := fmt.Sprintf("Error decoding base64 string: %v", err)
+		onUpdate(FirmwareUpdateMessage{FirmwareUpdateFailure: &msg})
+		handle.log.Error(msg)
+	}
+
+	onProgress := func(progressMsg string) {
+		onUpdate(FirmwareUpdateMessage{FirmwareUpdateProgress: &progressMsg})
+	}
+
+	err = firmware.Update(context.Background(), image, nil, &command.Address, onProgress)
```
knuton, line 48: A short wait before triggering the Update routine should help make sure TCP connection is torn down on both ends.

src/dividat-driver/firmware/main.go
```diff
@@ -100,35 +108,40 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
...
 			}
 		} else {
 			dfuHost = discoveredAddr
+			onProgress("Senso discovered via _sensoUpdate._udp")
 		}
 	}
 
+	onProgress("Preparing to transmit firmware.")
 	// 4: Transmit firmware via TFTP
 	time.Sleep(5 * time.Second) // Wait to ensure proper TFTP startup
```
knuton, line 133: This 5 seconds sleep might be a bit short if we are working with a `configuredAddr`. We get a TFTP timeout that looks like this:

image

Using 15 seconds worked for me to have the procedure work every time with a fixed address. Otherwise it only worked reliably when discovering the service instead of specifying a fixed address.

src/dividat-driver/firmware/main.go
```diff
@@ -100,35 +108,40 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 			// We do need to rediscover, as the legacy device may still just have
 			// restarted into the bootloader and obtained a new IP address.
 			ctx, cancel := context.WithTimeout(parentCtx, 60*time.Second)
-			legacyDiscoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx)
+			legacyDiscoveredAddr, err := discover("_sensoControl._tcp", deviceSerial, ctx, onProgress)
```
knuton, line 111: We could check the `mode` txt entry here to see if we discovered an actual bootloader.

src/dividat-driver/firmware/main.go
krksgbr: We can deduce the serial for fixed address configurations from the records returned in the discovery response.

krksgbr:
The latest changes should make the procedure a lot more reliable. It still doesn't seem bulletproof however. With this stress test I was able to send my controller into a state from which it can only be recovered by power cycling. After a few runs, the controller no longer announces itself via mdns and is also not responsive to commands. It's not typical usage, so I'm not sure how worried we should be about this, but I thought withstanding this would be a good indication of robustness. It's unclear if the conditions that bring this about are specific to the chained updates or if they can arise under normal use as well. In any case, the controller can be recovered by power cycling, so it's a relatively benign problem. We'll have to make sure to communicate this well in the UI. 

```diff
diff --git a/src/dividat-driver/firmware/main.go b/src/dividat-driver/firmware/main.go
index e0465d8..1ea09b6 100644
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -38,23 +38,30 @@ func Command(flags []string) {
 		flag.PrintDefaults()
 		return
 	}
-	file, err := os.Open(*imagePath)
-	if err != nil {
-		fmt.Printf("Could not open image file: %v\n", err)
-		os.Exit(1)
-	}
 
-	onProgress := func(progressMsg string) {
-		fmt.Println(progressMsg)
-	}
+	for i := 0; i < 10; i++ {
 
-	err = Update(context.Background(), file, deviceSerial, configuredAddr, onProgress)
+		file, err := os.Open(*imagePath)
+		if err != nil {
+			fmt.Printf("Could not open image file: %v\n", err)
+			os.Exit(1)
+		}
 
-	if err != nil {
-		fmt.Println(err.Error())
-		fmt.Println()
-		fmt.Println("Update failed. Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again.")
-		os.Exit(1)
+		onProgress := func(progressMsg string) {
+			fmt.Println(progressMsg)
+		}
+
+		fmt.Printf("Update run %d out of %d\n", i+1, 10)
+		err = Update(context.Background(), file, deviceSerial, configuredAddr, onProgress)
+		fmt.Printf("Pausing for 5 second\n")
+		time.Sleep(5 * time.Second)
+
+		if err != nil {
+			fmt.Println(err.Error())
+			// fmt.Println()
+			// fmt.Println("Update failed. Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again.")
+			// os.Exit(1)
+		}
 	}
 } 
```

src/dividat-driver/firmware/main.go
```diff
@@ -129,9 +129,9 @@ func Update(parentCtx context.Context, image io.Reader, deviceSerial *string, co
 		}
 	}
 
-	onProgress("Waiting 5 seconds to ensure proper TFTP startup")
+	onProgress("Waiting 15 seconds to ensure proper TFTP startup")
 	// 4: Transmit firmware via TFTP
-	time.Sleep(5 * time.Second) // Wait to ensure proper TFTP startup
```
knuton, line 120: Another thing we could do, if we want to optimize for time, is to add the extra wait only in the case where we use a `configuredAddr` for the `dfuHost`. Then we don't need to do extra waiting in the mDNS scenario.

krksgbr:
In an effort to address the flakiness observed in discovery, I switched from [grandcat/zeroconf](https://github.com/grandcat/zeroconf) to the more actively maintained fork, [libp2p/zeroconf](https://github.com/libp2p/zeroconf). This change inadvertently broke the build. I implemented a makeshift solution in https://github.com/dividat/driver/pull/127/commits/6da5b8bb37abf2a9b5d1d3d4c68e6393c4d0d4aa to restore the build, although cross-building is still broken. For now, I've left it unresolved, hoping we can use https://github.com/dividat/driver/pull/129.

I've also refactored and enhanced the discovery logic in https://github.com/dividat/driver/pull/127/commits/c7608d7ff76ea1cdbb47d48a1f56cf3597b9122a. It's unclear whether the improvements came from the refactor, the switch to the libp2p/zeroconf fork, or a combination of both, but discovery is noticeably more reliable now. Somehow, these adjustments also positively impacted other, seemingly unrelated aspects of the update process. The system can now complete all iterations of the previous stress test without failure.

<details><summary>New diff to stress test</summary>
<p>

```diff
diff --git a/src/dividat-driver/firmware/cli.go b/src/dividat-driver/firmware/cli.go
index 1e47801..bfa4edf 100644
--- a/src/dividat-driver/firmware/cli.go
+++ b/src/dividat-driver/firmware/cli.go
@@ -23,29 +23,34 @@ func Command(flags []string) {
 		flag.PrintDefaults()
 		return
 	}
-	file, err := os.Open(*imagePath)
-	if err != nil {
-		fmt.Printf("Could not open image file: %v\n", err)
-		os.Exit(1)
-	}
 
 	onProgress := func(progressMsg string) {
 		fmt.Println(progressMsg)
 	}
 
-	if *sensoSerial != "" {
-		err = UpdateBySerial(context.Background(), *sensoSerial, file, onProgress)
-	} else if *configuredAddr != "" {
-		err = updateByAddress(context.Background(), *configuredAddr, file, onProgress)
-	} else {
-		err = updateByDiscovery(context.Background(), file, onProgress)
-	}
-
-	if err != nil {
-		fmt.Println()
-		fmt.Printf("Update failed: %v \n", err)
-		os.Exit(1)
+	for i := 0; i < 15; i++ {
+		fmt.Println("Run ", i)
+		file, err := os.Open(*imagePath)
+		if err != nil {
+			fmt.Printf("Could not open image file: %v\n", err)
+			os.Exit(1)
+		}
+		if *sensoSerial != "" {
+			err = UpdateBySerial(context.Background(), *sensoSerial, file, onProgress)
+		} else if *configuredAddr != "" {
+			err = updateByAddress(context.Background(), *configuredAddr, file, onProgress)
+		} else {
+			err = updateByDiscovery(context.Background(), file, onProgress)
+		}
+		fmt.Println(err)
+		fmt.Println("Sleeping for 5s")
+		time.Sleep(5 * time.Second)
 	}
+	// if err != nil {
+	// 	fmt.Println()
+	// 	fmt.Printf("Update failed: %v \n", err)
+	// 	os.Exit(1)
+	// }
 }
 
 func updateByAddress(ctx context.Context, address string, image io.Reader, onProgress OnProgress) error {
```

</p>
</details>

src/dividat-driver/firmware/main.go
krksgbr: In [Refactor discovery code](https://github.com/dividat/driver/pull/127/commits/c7608d7ff76ea1cdbb47d48a1f56cf3597b9122a) I added support for getting the `mode=` txt entry. 

https://github.com/dividat/driver/blob/6da5b8bb37abf2a9b5d1d3d4c68e6393c4d0d4aa/src/dividat-driver/service/main.go#L142-L155

The `service.Find` call here spawns two goroutines looking for both `_sensoControl._tcp` and `_sensoUpdate._udp` and returns whichever is found first:
https://github.com/dividat/driver/blob/6da5b8bb37abf2a9b5d1d3d4c68e6393c4d0d4aa/src/dividat-driver/firmware/cli.go#L53

With this we can run a discovery first even for a configured address which allows us to learn right away if a Senso is in normal or bootloader mode (as well as validating the address). Then this timeout can just be part of the dfu sending logic:
https://github.com/dividat/driver/blob/6da5b8bb37abf2a9b5d1d3d4c68e6393c4d0d4aa/src/dividat-driver/firmware/main.go#L38-L68

src/dividat-driver/firmware/main.go
krksgbr: I increased the timeout in [Wait 15 seconds for tftp startup](https://github.com/dividat/driver/pull/127/commits/4201904cb5da062b37efd052668b3d871370b80c). However, I ran multiple stress tests with the latest changes and with only five seconds, and it seems to be enough, so I reduced the timeout back down to 5s. I wonder if this will prove to be robust enough under different network conditions though, so more testing is required.

src/dividat-driver/senso/update_firmware.go
krksgbr: I added a 10 second timeout in [Wait 10 seconds for tcp connection teardown](https://github.com/dividat/driver/pull/127/commits/94b819a57baf8bb70375236b5cba8590852c5102), but later  found this to be unnecessary, especially with [Send DFU with exponential backoff](https://github.com/dividat/driver/pull/127/commits/03b962d74561ee234be5e016be6fc3df5b698bd2).

knuton:
Should we tackle https://github.com/dividat/driver/pull/129 first then to unblock the newer Go libs?

src/dividat-driver/firmware/main.go
```diff
@@ -1,134 +1,97 @@
...
+
+		if dfuService == nil {
+			return fmt.Errorf("Could not rediscover Senso in bootloader mode.\n%s", tryPowerCycling)
 		}
+
+		target = *dfuService
+		onProgress(fmt.Sprintf("Re-discovered Senso in bootloader mode at %s", target.Address))
+		onProgress("Waiting 15 seconds to ensure proper TFTP startup")
+		// Wait to ensure proper TFTP startup
+		time.Sleep(5 * time.Second)
```
knuton, line 80: 5 or 15? Is a long/any wait even still needed with the retry-logic?

knuton (changes requested):
Looks very good!

Here are a bunch of initial suggestions a thorough reading of the code. I am going to start testing practical scenarios next.

knuton:
<details>
<summary>Tests with 60 second timeouts</summary>

```diff
diff --git a/src/dividat-driver/firmware/cli.go b/src/dividat-driver/firmware/cli.go
index 50a81a0..a30cefa 100644
--- a/src/dividat-driver/firmware/cli.go
+++ b/src/dividat-driver/firmware/cli.go
@@ -53,7 +53,7 @@ func Command(flags []string) {
 
 func updateByAddress(ctx context.Context, address string, image io.Reader, onProgress OnProgress) error {
 	onProgress(fmt.Sprintf("Using specified address %s", address))
-	match := service.Find(ctx, 15*time.Second, service.AddressFilter(address))
+	match := service.Find(ctx, 60*time.Second, service.AddressFilter(address))
 	if match == nil {
 		return fmt.Errorf("Failed to find Senso with address %s.\n%s", address, tryPowerCycling)
 	}
@@ -63,7 +63,7 @@ func updateByAddress(ctx context.Context, address string, image io.Reader, onPro
 
 func updateByDiscovery(ctx context.Context, image io.Reader, onProgress OnProgress) error {
 	onProgress("Discovering sensos")
-	services := service.List(ctx, 15*time.Second)
+	services := service.List(ctx, 60*time.Second)
 	if len(services) == 1 {
 		target := services[0]
 		onProgress(fmt.Sprintf("Discovered Senso: %s (%s)", target.Text.Serial, target.Address))
diff --git a/src/dividat-driver/firmware/main.go b/src/dividat-driver/firmware/main.go
index b7ee3cb..3926048 100644
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -37,7 +37,7 @@ const tryPowerCycling = "Try turning the Senso off and on, waiting for 30 second
 
 func UpdateBySerial(ctx context.Context, deviceSerial string, image io.Reader, onProgress OnProgress) error {
 	onProgress(fmt.Sprintf("Using specified serial %s", deviceSerial))
-	match := service.Find(ctx, 15*time.Second, service.SerialNumberFilter(deviceSerial))
+	match := service.Find(ctx, 60*time.Second, service.SerialNumberFilter(deviceSerial))
 	if match == nil {
 		return fmt.Errorf("Failed to find Senso with serial number %s.\n%s", deviceSerial, tryPowerCycling)
 	}
@@ -65,7 +65,7 @@ func update(parentCtx context.Context, target service.Service, image io.Reader,
 		}
 
 		onProgress("Looking for senso in bootloader mode")
-		dfuService := service.Find(parentCtx, 30*time.Second, func(discovered service.Service) bool {
+		dfuService := service.Find(parentCtx, 60*time.Second, func(discovered service.Service) bool {
 			return service.SerialNumberFilter(target.Text.Serial)(discovered) && service.IsDfuService(discovered)
 		})
 
```
</details>

41 consecutive runs until cancelled by me, when specifying serial on CLI:

    successful_runs=0; while bin/dividat-driver update-firmware -i ../senso-firmware/build/images/controller-app.bin -s 31-00000000; do ((successful_runs++)); echo "Runs: $successful_runs"; sleep 10; done

9 consecutive runs until the 10th failed, when not specifying serial on CLI:

    successful_runs=0; while bin/dividat-driver update-firmware -i ../senso-firmware/build/images/controller-app.bin; do ((successful_runs++)); echo "Runs: $successful_runs"; sleep 10; done

Output of failing run:

    Discovering sensos
    Discovered Senso: 31-00000000 (169.254.51.151)
    Sent DFU command to 169.254.51.151:55567.
    Looking for senso in bootloader mode
    
    Update failed: Could not rediscover Senso in bootloader mode.
    Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again. 

It seems like the Senso had somehow not entered bootloader mode, I restarted the loop and the first run looked like this:

    Discovering sensos
    Discovered Senso: 31-00000000 (169.254.51.151)
    Sent DFU command to 169.254.51.151:55567.
    Looking for senso in bootloader mode
    Re-discovered Senso in bootloader mode at 169.254.51.151
    Waiting 15 seconds to ensure proper TFTP startup
    Creating TFTP client
    Preparing transmission
    Failed on attempt 1, retrying in 2s
    Transmitting...
    119720 bytes sent
    Success! Firmware transmitted to Senso.

Following this, performed another loop with 46 successful consecutive runs without serial on CLI. So the one failure I observed may be unrelated to whether or not a serial was specified and was easily recoverable.

knuton (changes requested):

src/dividat-driver/firmware/main.go
```diff
@@ -1,134 +1,97 @@
...
+const tryPowerCycling = "Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again."
 
-	err = Update(context.Background(), file, deviceSerial, configuredAddr)
-	if err != nil {
-		fmt.Println(err.Error())
-		fmt.Println()
-		fmt.Println("Update failed. Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again.")
-		os.Exit(1)
+func UpdateBySerial(ctx context.Context, deviceSerial string, image io.Reader, onProgress OnProgress) error {
+	onProgress(fmt.Sprintf("Using specified serial %s", deviceSerial))
```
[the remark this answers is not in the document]
krksgbr: I think we could use "looking for" where we're looking for a specific Senso  - by address or by serial.

When we're looking for a specific device, it sounds a bit strange to me to say "Discovering Senso with serial/address", as "discovery" implies that we don't exactly know what we're looking for or what we are going to find.

When no address or serial is specified, we could use the word "discovering", though we could also say "Looking for Sensos" to be consistent, which I think still sounds fine in this context.

src/dividat-driver/firmware/main.go
```diff
@@ -1,134 +1,97 @@
...
+	match := service.Find(ctx, 15*time.Second, service.SerialNumberFilter(deviceSerial))
```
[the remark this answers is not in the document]
krksgbr: [Allow 60 seconds for mDNS discovery](https://github.com/dividat/driver/pull/127/commits/8097df36c246f9a1d700cbb5815e3f28acad5b90)

src/dividat-driver/firmware/main.go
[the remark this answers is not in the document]
knuton: Yeah, agree that the "looking for" phrasing is more natural overall. "Discovery" is the usual term in the context of mDNS, but I don't think we gain much from sticking to that term here, as readers are likely to fall into two camps: either they already know that mDNS is being used, or they would not know what mDNS is at all.

src/dividat-driver/firmware/main.go
krksgbr: The wait is still necessary even with retry-logic. I tried removing the timeout, but while testing consecutive updates, after a few runs `putTFTP` fails. Not only that, but the Senso doesn't respond anymore when running a new update, and it needs to be power cycled or sometimes even a re-flashed. I've found 5 seconds to work quite well, but I'm still observing occasional failure after many runs. I'm not sure if it's because this timeout is still too short. I'm experimenting with longer timeouts and see if it has any influence.

I'm not sure why this happens. It seems like when the Senso receives a TFTP write request too early after rebooting into bootloader mode, it gets into a weird state that it cannot recover from.

src/dividat-driver/firmware/cli.go
```diff
@@ -0,0 +1,76 @@
...
+		os.Exit(1)
+	}
+}
+
+// The following functions are only used when updating the firmware via the command line.
+// This is why they are private, and not part of the main module.
+
+func updateByAddress(ctx context.Context, address string, image io.Reader, onProgress OnProgress) error {
+	onProgress(fmt.Sprintf("Using specified address %s", address))
+	match := service.Find(ctx, 15*time.Second, service.AddressFilter(address))
```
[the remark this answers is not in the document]
krksgbr: You're right, this slipped my attention. After browsing some sysadmin forums I get the impression it's not uncommon to block mDNS traffic. Since we allow people to configure Play to use a fixed address, I think we should extend this to the firmware upgrade as well. We could make `updateByAddress` assume that it's only used where mDNS is not available (ie. when Play is configured with a static address):

```go
func updateByAddress(ctx context.Context, address string, image io.Reader, onProgress OnProgress) error {
	onProgress(fmt.Sprintf("Using specified address %s", address))
	err := sendDfuCommand(address, controllerPort, onProgress)
	if err == nil {
		waitForTFTP(onProgress)
	} else {
		onProgress("Could not send DFU command. Senso may already be in bootloader mode.")
	}
	return putTFTP(address, tftpPort, image, onProgress)
}
```

This requires that a static DHCP lease is configured for the Senso, otherwise the Senso could potentially get stuck in bootloader mode if its address changes when it reboots. Hopefully clients with a static address configuration would set this up as well.

src/dividat-driver/firmware/main.go
krksgbr: Increasing the wait to 10 seconds improved robustness. I was able to do over 250 back-to-back updates before killing the test script.

[Increase wait time before attempting TFTP transfer](https://github.com/dividat/driver/pull/127/commits/ffa7b584a1a33fedbb92dad271044cc829518aee)

src/dividat-driver/firmware/cli.go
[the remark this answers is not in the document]
knuton: > This requires that a static DHCP lease is configured for the Senso, otherwise the Senso could potentially get stuck in bootloader mode if its address changes when it reboots. Hopefully clients with a static address configuration would set this up as well.

Yes, we need to assume a statically configured address in this case. A Senso in normal mode could self-assign a new address each time, if no DHCP lease is given. So there should be a stable assignment mechanism, usually based on MAC.

> Since we allow people to configure Play to use a fixed address, I think we should extend this to the firmware upgrade as well.

This reasoning makes sense, but I think it equally makes sense to only support the mDNS case for firmware updates, because

- it is used by 99 % of installations,
- it helps us provide a robust update mechanism through better insight into peer state,
- we can focus our resources on making this one mechanism work well.

It would still be reasonable to say "normal operation is supported without mDNS, for updating attach Senso directly to Play Computer".

src/dividat-driver/firmware/main.go
knuton: :+1: 

23 runs until cancelled by me:

    successful_runs=0; while bin/dividat-driver update-firmware -i ../senso-firmware/build/images/controller-app.bin; do ((successful_runs++)); echo "Runs: $successful_runs"; sleep 1; done

Good to go with this

src/dividat-driver/senso/websocket.go
```diff
@@ -233,6 +286,11 @@ func (handle *Handle) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				}
 				log.WithField("command", prettyPrintCommand(command)).Debug("Received command.")
 
+				if handle.firmwareUpdate.IsUpdating() && (command.GetStatus == nil || command.Discover == nil) {
+					handle.log.Info("Firmware update in progress, ignoring UpdateFirmware command.")
```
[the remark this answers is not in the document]
krksgbr: Yes, this was incorrect. Fixed it in:
[Improve log messages about ignored commands during firmware update](https://github.com/dividat/driver/pull/127/commits/0121d02a32cb56394d6aa5be7c916d2ad5bea555)

Also related:
[Implement pretty-printing for UpdateFirmware command](https://github.com/dividat/driver/pull/127/commits/80dfde6851a9755d618894d995ffad8e15022a32)

I'm wondering about the log level. Is `Info` appropriate? I'm thinking maybe `Debug` would be enough.

src/dividat-driver/firmware/main.go
```diff
@@ -1,134 +1,97 @@
...
-	if *imagePath == "" {
-		flag.PrintDefaults()
-		return
-	}
-	file, err := os.Open(*imagePath)
-	if err != nil {
-		fmt.Printf("Could not open image file: %v\n", err)
-		os.Exit(1)
-	}
+const tryPowerCycling = "Try turning the Senso off and on, waiting for 30 seconds and then running this update tool again."
```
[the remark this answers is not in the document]
krksgbr: Agree:
[Communicate power cycling suggestion only when updating via CLI](https://github.com/dividat/driver/pull/127/commits/cb649fcaf64a7bb0e40068584cb841cc02421e7d)

src/dividat-driver/firmware/cli.go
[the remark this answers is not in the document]
krksgbr: The implementation of updating by address seems quite straightforward to me so I was tempted to keep it, but seeing that the overwhelming majority of people uses mDNS, I'm convinced to only offer updating via serial for now. Let's see how well this one mechanism will work under real-world conditions. We can always re-add updating by address if it proves to be relevant, but it would be harder to remove later if some people come to rely on it.

[Remove firmware update by fixed address](https://github.com/dividat/driver/pull/127/commits/8743fda024225174829514e81639e00105dced06)

go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/ebfe/scard v0.0.0-20190212122703-c3d1b1916a95
 	github.com/gorilla/websocket v1.4.2
-	github.com/grandcat/zeroconf v1.0.0
 	github.com/kardianos/service v1.2.0
+	github.com/libp2p/zeroconf/v2 v2.2.0
```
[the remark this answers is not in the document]
krksgbr: I didn't try a newer version of `grandcat/zeroconf`.
We could consider trying as well, though the `libp2p` version seems to work quite well. I've noticed some improvements in robustness of mDNS discovery since switching to it, though I'm not sure if it's because of the changes in the library or refactoring our own code.

There are also quite a few projects using `grandcat/zeroconf`, some of them are quite popular:
https://github.com/grandcat/zeroconf/network/dependents

It probably does have a larger community around it.

Looking at the difference between `grandcat/zeroconf` and `libp2p/zeroconf`:

https://github.com/libp2p/zeroconf/compare/master...grandcat%3Azeroconf%3Amaster
https://github.com/grandcat/zeroconf/compare/master...libp2p:zeroconf:master

There's not a lot in `grandcat/zeroconf` that's not in `libp2p/zeroconf`, but the latter seems to contain quite a few fixes. It's also easier to use because of an API change described [here](https://github.com/grandcat/zeroconf/issues/98), which was implemented in the libp2p version.

src/dividat-driver/firmware/main.go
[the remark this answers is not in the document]
krksgbr: [Improve clarity of firmware update message](https://github.com/dividat/driver/pull/127/commits/ccb18c5430ec4ed65fff79aaeb5dfcf6131a1eab)

go.mod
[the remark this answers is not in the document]
krksgbr: I tested a newer version of `grandcat/zeroconf` and it seems to work just as well.

src/dividat-driver/senso/websocket.go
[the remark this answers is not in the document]
knuton: Agree, `Debug` should be a good level to log this at.

go.mod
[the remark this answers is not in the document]
krksgbr: I'd say let's stick with the fork since we've been testing with it and we know it works well.  I'll add a comment to `go.mod`.

src/dividat-driver/senso/websocket.go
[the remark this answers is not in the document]
krksgbr: [Log messages about ignored commands at debug level](https://github.com/dividat/driver/pull/127/commits/81d9acd5cab149d4f6bec11a0c71999f4aa8d1b6)

go.mod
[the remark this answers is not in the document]
krksgbr: [Add notes about libp2p/zeroconf dependency choice](https://github.com/dividat/driver/pull/127/commits/ad56fd209c664e2b0fc8ea9fb71bcb06ada10d17)

src/dividat-driver/senso/websocket.go
```diff
@@ -95,15 +110,21 @@ func (command *Command) UnmarshalJSON(data []byte) error {
...
 type Status struct {
 	Address *string
 }
 
+type FirmwareUpdateMessage struct {
+	FirmwareUpdateProgress *string
+	FirmwareUpdateSuccess  *string
+	FirmwareUpdateFailure  *string
+}
+
```
krksgbr, line 127: I'm still wondering about whether to simplify this into something like:

```go
type FirmwareUpdateMessage struct {
	Type    string
	Message string
}
```

It could simplify code both here and the client. A potential downside of merging the three messages is that it would make it more difficult to evolve them independently, eg. add some fields to some of them but not the others.

krksgbr (commented):
There's one more detail I'm undecided about. What are your thoughts on this @knuton?

src/dividat-driver/senso/websocket.go
krksgbr: I quickly tried the alternative I came to the conclusion that the representation is fine as it is, but I'd simplify json serialization to remove a bunch of code:

```diff
diff --git a/src/dividat-driver/senso/websocket.go b/src/dividat-driver/senso/websocket.go
index ce19d83..763e18d 100644
--- a/src/dividat-driver/senso/websocket.go
+++ b/src/dividat-driver/senso/websocket.go
@@ -110,8 +110,8 @@ func (command *Command) UnmarshalJSON(data []byte) error {
 // Message that can be sent to Play
 type Message struct {
 	*Status
-	Discovered            *zeroconf.ServiceEntry
-	FirmwareUpdateMessage *FirmwareUpdateMessage
+	Discovered             *zeroconf.ServiceEntry
+	FirmwareUpdateResponse *FirmwareUpdateMessage
 }
 
 // Status is a message containing status information
@@ -119,10 +119,10 @@ type Status struct {
 	Address *string
 }
 
+// Message to update client about firmware update progress
 type FirmwareUpdateMessage struct {
-	FirmwareUpdateProgress *string
-	FirmwareUpdateSuccess  *string
-	FirmwareUpdateFailure  *string
+	Type    string `json:"type"`
+	Message string `json:"message"`
 }
 
 // MarshalJSON ipmlements JSON encoder for messages
@@ -147,34 +147,8 @@ func (message *Message) MarshalJSON() ([]byte, error) {
 			IP:           append(message.Discovered.AddrIPv4, message.Discovered.AddrIPv6...),
 		})
 
-	} else if message.FirmwareUpdateMessage != nil {
-		fwUpdate := struct {
-			Type    string `json:"type"`
-			Message string `json:"message"`
-		}{}
-
-		firmwareUpdateMessage := *message.FirmwareUpdateMessage
-
-		if firmwareUpdateMessage.FirmwareUpdateProgress != nil {
-
-			fwUpdate.Type = "FirmwareUpdateProgress"
-			fwUpdate.Message = *firmwareUpdateMessage.FirmwareUpdateProgress
-
-		} else if firmwareUpdateMessage.FirmwareUpdateFailure != nil {
-
-			fwUpdate.Type = "FirmwareUpdateFailure"
-			fwUpdate.Message = *firmwareUpdateMessage.FirmwareUpdateFailure
-
-		} else if firmwareUpdateMessage.FirmwareUpdateSuccess != nil {
-
-			fwUpdate.Type = "FirmwareUpdateSuccess"
-			fwUpdate.Message = *firmwareUpdateMessage.FirmwareUpdateSuccess
-
-		} else {
-			return nil, errors.New("could not marshal firmware update message")
-		}
-
-		return json.Marshal(fwUpdate)
+	} else if message.FirmwareUpdateResponse != nil {
+		return json.Marshal(message.FirmwareUpdateResponse)
 	}
 
 	return nil, errors.New("could not marshal message")
@@ -356,33 +330,21 @@ func (handle *Handle) dispatchCommand(ctx context.Context, log *logrus.Entry, co
 	} else if command.UpdateFirmware != nil {
 		go handle.ProcessFirmwareUpdateRequest(*command.UpdateFirmware, SendMsg{
 			progress: func(msg string) {
-				sendMessage(firmwareUpdateProgress(msg))
+				sendMessage(firmwareUpdateMessage("FirmwareUpdateProgress", msg))
 			},
 			failure: func(msg string) {
-				sendMessage(firmwareUpdateFailure(msg))
+				sendMessage(firmwareUpdateMessage("FirmwareUpdateFailure", msg))
 			},
 			success: func(msg string) {
-				sendMessage(firmwareUpdateSuccess(msg))
+				sendMessage(firmwareUpdateMessage("FirmwareUpdateSuccess", msg))
 			},
 		})
 	}
 	return nil
 }
 
-func firmwareUpdateSuccess(msg string) Message {
-	return firmwareUpdateMessage(FirmwareUpdateMessage{FirmwareUpdateSuccess: &msg})
-}
-
-func firmwareUpdateFailure(msg string) Message {
-	return firmwareUpdateMessage(FirmwareUpdateMessage{FirmwareUpdateFailure: &msg})
-}
-
-func firmwareUpdateProgress(msg string) Message {
-	return firmwareUpdateMessage(FirmwareUpdateMessage{FirmwareUpdateProgress: &msg})
-}
-
-func firmwareUpdateMessage(msg FirmwareUpdateMessage) Message {
-	return Message{FirmwareUpdateMessage: &msg}
+func firmwareUpdateMessage(responseType string, msg string) Message {
+	return Message{FirmwareUpdateResponse: &FirmwareUpdateMessage{Type: responseType, Message: msg}}
 }
 
 // rx_data_loop reads data from Senso and forwards it up the WebSocket

```

However this would introduce some inconsistency to how firmware update messages vs all the other messages are serialized, so I'd rather keep things as they are now. I think it would be worth refactoring, but it would be better to change all of the messages at once to follow the same pattern.

knuton: I wonder whether the same pattern would work for the others, seeing that unlike with these three, each variant carries different fields?

Maybe type switches could be useful? https://go.dev/tour/methods/16

knuton (changes requested):
All very good, just some nitpicks on the progress messages, sorry! :upside_down_face: 

I think the output is going to be easier to understand if we use uniform verbiage, such as "looking for", "found"/"not found", without also speaking about "discovery".

I believe the "to discover" verb just comes from mDNS/service discovery, and only makes sense when viewed from this technical angle. Otherwise it seems quite weird to me in English to say "I am trying to discover the lawnmower." or "I looked everywhere, but I could not rediscover my keys!".

Less importantly, would be nice to be consistent with whether or not messages end in periods.

src/dividat-driver/firmware/main.go
```diff
@@ -158,97 +119,76 @@ func sendDfuCommand(host string, port string) error {
 		return fmt.Errorf("Could not send DFU command: %v", err)
 	}
 
-	fmt.Printf("Sent DFU command to %s:%s.\n", host, port)
+	onProgress(fmt.Sprintf("Sent DFU command to %s:%s.", host, port))
```
krksgbr, line 122:
```suggestion
	onProgress(fmt.Sprintf("Sent DFU command to %s:%s", host, port))
```

src/dividat-driver/senso/update_firmware.go
```diff
@@ -0,0 +1,54 @@
...
+		return
+	}
+
+	err = firmware.UpdateBySerial(context.Background(), command.SerialNumber, image, send.progress)
+	if err != nil {
+		failureMsg := fmt.Sprintf("Failed to update firmware: %v", err)
+		send.failure(failureMsg)
+		handle.log.Error(failureMsg)
+	} else {
+		send.success("Firmware successfully transmitted.")
```
krksgbr, line 43:
```suggestion
		send.success("Firmware successfully transmitted")
```

krksgbr (commented):
Adding a couple more refinements to the batch, removing periods

src/dividat-driver/senso/websocket.go
krksgbr: I think something like this could work.

<details><summary>Diff</summary>
<p>

```diff
diff --git a/src/dividat-driver/senso/messages.go b/src/dividat-driver/senso/messages.go
new file mode 100644
index 0000000..2149687
--- /dev/null
+++ b/src/dividat-driver/senso/messages.go
@@ -0,0 +1,64 @@
+package senso
+
+import (
+	"encoding/json"
+	"net"
+
+	"github.com/libp2p/zeroconf/v2"
+)
+
+type Message struct {
+	*Status
+	*Discovered
+	*FirmwareUpdateResponse
+	Type string `json:"type"`
+}
+
+// Status
+
+type Status struct {
+	Address *string `json:"address"`
+}
+
+func (s *Status) MarshalJSON() ([]byte, error) {
+	return json.Marshal(Message{
+		Type:   "Status",
+		Status: s,
+	})
+}
+
+// Discovered
+
+type Discovered struct {
+	ServiceEntry *zeroconf.ServiceEntry `json:"service"`
+	IP           []net.IP               `json:"ip"`
+}
+
+func (d *Discovered) MarshalJSON() ([]byte, error) {
+	return json.Marshal(Message{
+		Type:       "Discovered",
+		Discovered: d,
+	})
+}
+
+// Firmware update messages
+
+type FirmwareUpdateResponseType string
+
+const (
+	FirmwareUpdateSuccess  FirmwareUpdateResponseType = "FirmwareUpdateSuccess"
+	FirmwareUpdateFailure  FirmwareUpdateResponseType = "FirmwareUpdateFailure"
+	FirmwareUpdateProgress FirmwareUpdateResponseType = "FirmwareUpdateProgress"
+)
+
+type FirmwareUpdateResponse struct {
+	Type    FirmwareUpdateResponseType
+	Message string `json:"message"`
+}
+
+func (r *FirmwareUpdateResponse) MarshalJSON() ([]byte, error) {
+	return json.Marshal(Message{
+		Type:                   string(r.Type),
+		FirmwareUpdateResponse: r,
+	})
+}
diff --git a/src/dividat-driver/senso/websocket.go b/src/dividat-driver/senso/websocket.go
index ce19d83..eb7ba53 100644
--- a/src/dividat-driver/senso/websocket.go
+++ b/src/dividat-driver/senso/websocket.go
@@ -4,13 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"net"
 	"net/http"
 	"sync"
 	"time"
 
 	"github.com/gorilla/websocket"
-	"github.com/libp2p/zeroconf/v2"
 	"github.com/sirupsen/logrus"
 
 	"github.com/dividat/driver/src/dividat-driver/service"
@@ -107,80 +105,6 @@ func (command *Command) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// Message that can be sent to Play
-type Message struct {
-	*Status
-	Discovered            *zeroconf.ServiceEntry
-	FirmwareUpdateMessage *FirmwareUpdateMessage
-}
-
-// Status is a message containing status information
-type Status struct {
-	Address *string
-}
-
-type FirmwareUpdateMessage struct {
-	FirmwareUpdateProgress *string
-	FirmwareUpdateSuccess  *string
-	FirmwareUpdateFailure  *string
-}
-
-// MarshalJSON ipmlements JSON encoder for messages
-func (message *Message) MarshalJSON() ([]byte, error) {
-	if message.Status != nil {
-		return json.Marshal(&struct {
-			Type    string  `json:"type"`
-			Address *string `json:"address"`
-		}{
-			Type:    "Status",
-			Address: message.Status.Address,
-		})
-
-	} else if message.Discovered != nil {
-		return json.Marshal(&struct {
-			Type         string                 `json:"type"`
-			ServiceEntry *zeroconf.ServiceEntry `json:"service"`
-			IP           []net.IP               `json:"ip"`
-		}{
-			Type:         "Discovered",
-			ServiceEntry: message.Discovered,
-			IP:           append(message.Discovered.AddrIPv4, message.Discovered.AddrIPv6...),
-		})
-
-	} else if message.FirmwareUpdateMessage != nil {
-		fwUpdate := struct {
-			Type    string `json:"type"`
-			Message string `json:"message"`
-		}{}
-
-		firmwareUpdateMessage := *message.FirmwareUpdateMessage
-
-		if firmwareUpdateMessage.FirmwareUpdateProgress != nil {
-
-			fwUpdate.Type = "FirmwareUpdateProgress"
-			fwUpdate.Message = *firmwareUpdateMessage.FirmwareUpdateProgress
-
-		} else if firmwareUpdateMessage.FirmwareUpdateFailure != nil {
-
-			fwUpdate.Type = "FirmwareUpdateFailure"
-			fwUpdate.Message = *firmwareUpdateMessage.FirmwareUpdateFailure
-
-		} else if firmwareUpdateMessage.FirmwareUpdateSuccess != nil {
-
-			fwUpdate.Type = "FirmwareUpdateSuccess"
-			fwUpdate.Message = *firmwareUpdateMessage.FirmwareUpdateSuccess
-
-		} else {
-			return nil, errors.New("could not marshal firmware update message")
-		}
-
-		return json.Marshal(fwUpdate)
-	}
-
-	return nil, errors.New("could not marshal message")
-
-}
-
 // Implement net/http Handler interface
 func (handle *Handle) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
@@ -222,7 +146,7 @@ func (handle *Handle) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// send messgae up the WebSocket
-	sendMessage := func(message Message) error {
+	sendMessage := func(message json.Marshaler) error {
 		writeMutex.Lock()
 		conn.SetWriteDeadline(time.Now().Add(50 * time.Millisecond))
 		err := conn.WriteJSON(&message)
@@ -307,15 +231,13 @@ func (handle *Handle) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // HELPERS
 
 // dispatchCommand handles incomming commands and sends responses back up the WebSocket
-func (handle *Handle) dispatchCommand(ctx context.Context, log *logrus.Entry, command Command, sendMessage func(Message) error) error {
+func (handle *Handle) dispatchCommand(ctx context.Context, log *logrus.Entry, command Command, sendMessage func(json.Marshaler) error) error {
 
 	if command.GetStatus != nil {
 
-		var message Message
-
-		message.Status = &Status{Address: handle.Address}
-
-		err := sendMessage(message)
+		err := sendMessage(&Status{
+			Address: handle.Address,
+		})
 
 		if err != nil {
 			return err
@@ -339,10 +261,11 @@ func (handle *Handle) dispatchCommand(ctx context.Context, log *logrus.Entry, co
 			for entry := range entries {
 				log.WithField("service", entry).Debug("Discovered service.")
 
-				var message Message
-				message.Discovered = &entry.ServiceEntry
-
-				err := sendMessage(message)
+				serviceEntry := &entry.ServiceEntry
+				err := sendMessage(&Discovered{
+					ServiceEntry: serviceEntry,
+					IP:           append(serviceEntry.AddrIPv4, serviceEntry.AddrIPv6...),
+				})
 				if err != nil {
 					return
 				}
@@ -356,35 +279,19 @@ func (handle *Handle) dispatchCommand(ctx context.Context, log *logrus.Entry, co
 	} else if command.UpdateFirmware != nil {
 		go handle.ProcessFirmwareUpdateRequest(*command.UpdateFirmware, SendMsg{
 			progress: func(msg string) {
-				sendMessage(firmwareUpdateProgress(msg))
+				sendMessage(&FirmwareUpdateResponse{Type: FirmwareUpdateProgress, Message: msg})
 			},
 			failure: func(msg string) {
-				sendMessage(firmwareUpdateFailure(msg))
+				sendMessage(&FirmwareUpdateResponse{Type: FirmwareUpdateFailure, Message: msg})
 			},
 			success: func(msg string) {
-				sendMessage(firmwareUpdateSuccess(msg))
+				sendMessage(&FirmwareUpdateResponse{Type: FirmwareUpdateSuccess, Message: msg})
 			},
 		})
 	}
 	return nil
 }
 
-func firmwareUpdateSuccess(msg string) Message {
-	return firmwareUpdateMessage(FirmwareUpdateMessage{FirmwareUpdateSuccess: &msg})
-}
-
-func firmwareUpdateFailure(msg string) Message {
-	return firmwareUpdateMessage(FirmwareUpdateMessage{FirmwareUpdateFailure: &msg})
-}
-
-func firmwareUpdateProgress(msg string) Message {
-	return firmwareUpdateMessage(FirmwareUpdateMessage{FirmwareUpdateProgress: &msg})
-}
-
-func firmwareUpdateMessage(msg FirmwareUpdateMessage) Message {
-	return Message{FirmwareUpdateMessage: &msg}
-}
-
 // rx_data_loop reads data from Senso and forwards it up the WebSocket
 func rx_data_loop(ctx context.Context, rx chan interface{}, send func([]byte) error) {
 	var err error

```

</p>
</details>

knuton (approved):
:vulcan_salute:

PR #129, opened 2024-03-28
Setup dev shell and cross compilation with nix flakes

krksgbr:
Builds on:
https://github.com/knuton/driver/commits/build-env-update-nixpkgs/

Context:
I'm working on MacOS and the current nix-shell is broken. I've been using a globally installed go to build, but somehow it no longer finds the PCSC headers when I run make. I'm not sure what changed. Thought I'd use the opportunity to upgrade the our nix setup. With this flake configuration I'm able to cross-compile for the systems specified in the flake file.
Using `gnu` for the linux build instead of `musl` based on [Maybe we could build for GNU/dynlinking instead?](https://github.com/dividat/driver/commit/f49df1708e2f537b15bb1f6db3e36656f26ddc09), but I've found that it also builds with `musl`. ~~What was the reason for the change? Also, do we need to customize the pcsc build like done in [Disable more pcsclite features](https://github.com/dividat/driver/commit/7e4d8d1b26665698333e2ff7b58a55763fe88375), or was that just to make the build succeed?~~ Edit: I've just read the commit messages and saw that this was just to restore the build. Should we stick to `musl` then?
 
## Checklist

-   [X] Changelog updated
-   [X] Code documented


knuton:
> I've just read the commit messages and saw that this was just to restore the build. Should we stick to `musl` then?

We [don't currently cross compile in PlayOS](https://github.com/dividat/playos/blob/main/application/overlays/dividat-driver/default.nix), so we don't _really_ need the Linux crossbuild right now. If it is easy to make work with musl, let's stick to that.

krksgbr:
> > I've just read the commit messages and saw that this was just to restore the build. Should we stick to `musl` then?
> 
> We [don't currently cross compile in PlayOS](https://github.com/dividat/playos/blob/main/application/overlays/dividat-driver/default.nix), so we don't _really_ need the Linux crossbuild right now. If it is easy to make work with musl, let's stick to that.

It was easy so I kept it. I also realized we actually couldn't have used `gnu` because when we build with `nix` (or a nix-built `go` with dynamic linking, the links will point to nix store paths, so they won't be usable on non-nix systems:

```
❯ ldd bin/dividat-driver
        linux-vdso.so.1 (0x0000ffffbdc6f000)
        libresolv.so.2 => /nix/store/cv8mfy5wdfwfw4iwhdlkl4ddy8apl667-glibc-2.38-27/lib/libresolv.so.2 (0x0000ffffbdc00000)
        libpthread.so.0 => /nix/store/cv8mfy5wdfwfw4iwhdlkl4ddy8apl667-glibc-2.38-27/lib/libpthread.so.0 (0x0000ffffbdbd0000)
        libpcsclite.so.1 => /nix/store/f698bqhkqmx9nn1j8pkrg73vvfsb9c9y-pcsclite-1.9.5/lib/libpcsclite.so.1 (0x0000ffffbdba0000)
        libc.so.6 => /nix/store/cv8mfy5wdfwfw4iwhdlkl4ddy8apl667-glibc-2.38-27/lib/libc.so.6 (0x0000ffffbd9f0000)
        /nix/store/cv8mfy5wdfwfw4iwhdlkl4ddy8apl667-glibc-2.38-27/lib/ld-linux-aarch64.so.1 (0x0000ffffbdc32000)
        librt.so.1 => /nix/store/cv8mfy5wdfwfw4iwhdlkl4ddy8apl667-glibc-2.38-27/lib/librt.so.1 (0x0000ffffbd9c0000)
```

We didn't have a darwin build before, but this was also trivial to add so I did. This uses dynamic linking, but the links point to standard locations, so I think it should work:
```
❯ otool -L ./result/bin/dividat-driver
./result/bin/dividat-driver:
        /usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
        /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1770.255.0)
        /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
        /System/Library/Frameworks/PCSC.framework/Versions/A/PCSC (compatibility version 0.0.0, current version 487.60.2)
        /System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 59754.60.13)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
```

krksgbr:
Still need to document stuff, but it's ready for initial review/testing.
I tried the cross-built executables on all platforms and they all run, but I didn't try actually using them.

knuton:
> We didn't have a darwin build before, but this was also trivial to add so I did. This uses dynamic linking, but the links point to standard locations, so I think it should work:

That's great! We had it before IIRC, but kicked it out when it brought complications and we didn't need it at the time (we offered [this Homebrew formula](https://github.com/dividat/homebrew-driver) for gamedevs on Mac).

krksgbr:
> > We didn't have a darwin build before, but this was also trivial to add so I did. This uses dynamic linking, but the links point to standard locations, so I think it should work:
> 
> That's great! We had it before IIRC, but kicked it out when it brought complications and we didn't need it at the time (we offered [this Homebrew formula](https://github.com/dividat/homebrew-driver) for gamedevs on Mac).

I should have qualified
> this was also trivial to add so I did

... with the caveat that building only works on macOS and that even there, cross builds between architectures is not yet supported when going through nix. Given these limitations, I think it's more sensible to cross compile directly with `go` and the system's `clang`. This enables us to build both for arm and amd: [Provide shell script for darwin cross builds](https://github.com/dividat/driver/pull/129/commits/277a08095a05daa5d8870253a76eeef638e0a1d2).


Github offers hosted runners, including [macOS ones](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), which we could look into when it becomes more relevant.


We could also consider [merging the two binaries](https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary) into a universal one which would run on both architectures. It's quite simple to do:
```
driver/bin | nix-flake* | ❄
❯ lipo -create -output dividat-driver-universal dividat-driver-amd64 dividat-driver-arm64

driver/bin | nix-flake* | ❄
❯ lipo -info dividat-driver-universal
Architectures in the fat file: dividat-driver-universal are: x86_64 arm64
```
This doubles the size from 10 to 20MB but it's still pretty small.

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,6 @@
+steps:
```
knuton, line 1: Did we previously define this in the Buildkite CI? Have you updated Buildkite to expect a YAML definition in the repo?

Or where did the steps previously come from?

Makefile
```diff
@@ -65,16 +55,31 @@ record:
...
+.PHONY: bin
+bin:
+	mkdir -p bin
+
+
+VERSION := $(shell git describe --tags --abbrev=0)
+
+.PHONY: VERSION
+VERSION:
+	echo $(VERSION) > VERSION
```
knuton, line 67: Does this need to be a file instead of a value given as a parameter?

Makefile
```diff
@@ -65,16 +55,31 @@ record:
 record-flex:
 	@go run src/dividat-driver/recorder/main.go ws://localhost:8382/flex
 
+.PHONY: bin
+bin:
+	mkdir -p bin
+
+
+VERSION := $(shell git describe --tags --abbrev=0)
```
knuton, line 63: Can you explain the intent here? It doesn't seem desirable to me at first sight, don't we want to know the actual version, instead of treating any commit following the latest tag as the same version?

flake.nix
```diff
@@ -0,0 +1,31 @@
+{
+  description = "Cross compile dividat driver";
```
knuton, line 2: nitpick: This does more than cross compile, provides dev environment, etc.

VERSION
```diff
@@ -0,0 +1 @@
+2.3.0
```
knuton, line 1: If we needed this file, we would probably still add it to `.gitignore`?

nix/crossBuild.nix
```diff
@@ -0,0 +1,85 @@
...
+      crossPkgs = (import nixpkgs {
+        system = system;
+        crossSystem.config = targetSystem;
+      });
+    in
+    with crossPkgs;
+    buildGoModule ({
+      pname = "dividat-driver";
+      version = version;
+      vendorHash = "sha256-S8qJorySFPERNp4i9ckeZCPLyx0d/QweS4hm4ghJT4k=";
```
knuton, line 18: Leave short note on how to update this?

nix/crossBuild.nix
```diff
@@ -0,0 +1,85 @@
...
+      ldflags = [
+        ldEmbedVersion
+        "-linkmode external"
+        "-extldflags '-static'"
```
knuton, line 22: Is this Linux-specific or does it apply to Windows too?

nix/crossBuild.nix
```diff
@@ -0,0 +1,85 @@
...
+      ];
+      src = ../.;
+      meta.platforms =
+        [
+          "x86_64-linux"
+          "x86_64-windows"
```
knuton, line 28: Not sure it matters, but should it always just be the one for the system actually being built for?

nix/devShell.nix
```diff
@@ -0,0 +1,38 @@
...
+    # Script to cross build on MacOS.
+    darwinCrossBuildScript
+
+  ]
+  ++ lib.optional stdenv.isLinux pcsclite
+  ++ lib.optional stdenv.isDarwin
+    # PCSC on Darwin
+    # This is only used for development and has nothing to do with `darwinCrossBuildScript`.
+    # The script uses the system's clang and PCSC framework, instead of gcc and nix's PCSC.
+    # See `nix/crossBuild.nix` for details.
```
knuton, line 35: From this comment I am not sure: Does it have to do with crossbuilding or not?

nix/crossBuild.nix
```diff
@@ -0,0 +1,85 @@
...
+      text =
+        if stdenv.isDarwin then ''
+          export GOOS=darwin
+          export CGO_ENABLED=1
+          export CC="/usr/bin/clang"
+
+          mkdir -p bin
+
+          ${buildCmd "amd64"}
+          ${buildCmd "arm64"}
```
knuton, line 77: I wonder how portable this is across macOS versions, do you know?

Could you share binaries built for AMD64?

knuton (commented):
Cool, thanks for driving this modernization!

I would suggest to focus on Linux and Windows here and tackle macOS only to the extent that it unblocks your feature development.

It's important to be able to build for macOS for you. Doing it as part of a crossbuild mechanism only really becomes useful to us however when we also provide a way for every team member to run tests, trigger builds, create releases for all OS. If I understand right your idea is that we could make crossbuilds and provide prebuilt binaries via GitHub Actions runners on macOS. This may be the best approach, but I'd spin it out as a future project.

nix/crossBuild.nix
krksgbr: Kind of an outdated answer, but this suggests static linking is also possible for windows builds:
https://stackoverflow.com/questions/43311621/c-correct-way-to-statically-dynamically-link-with-mingw-w64

VERSION
krksgbr: Let's discuss in this thread also the other questions surrounding this file.

> Does this need to be a file instead of a value given as a parameter?

It has to be a file, because nix flakes are [evaluated hermetically](https://www.tweag.io/blog/2020-07-31-nixos-flakes/#hermetic-evaluation), so we can't pass the version a parameter. We can't add the file to `.gitignore` either, because files used in a flake builds must be part of the working tree. Some alternatives I thought of:

1. Use the flake only to provide a shell with dependencies in which the binaries can be built, then it doesn't need to know about the version. I tried this initially, but ran into problems with some dependencies not being found.
2. We could make the flake be the source of truth in regards to the version, and expose it as an environment variable in the devshell. We could even include the git revision in the version string (almost like what `git describe` gives us):
```nix
      rev = if self ? shortRev then self.shortRev else "unknown";
      version = "2.3.0-${rev}";
```
This is only possible if the git tree is clean, which I guess is a good thing. Building with uncommitted changes, and using the last revision would not really be honest.

I'm wondering however if we even need the revision for the versions we release? As I understand, the binaries that are cross-built with nix are meant for deploying new releases. Wouldn't we make new releases from specific tags? In this case adding the revision seems redundant to me. Or do we also want to deploy releases from non-tagged commits?

.buildkite/pipeline.yml
krksgbr: I updated buildkite to expect a yaml definition. The steps were previously defined through the legacy (pre-yaml) step definition UI.

nix/crossBuild.nix
krksgbr: I tried the amd64 binary on my older mac running 10.15.7 and it worked.

[dividat-driver-amd64.tar.gz](https://github.com/dividat/driver/files/14897255/dividat-driver-amd64.tar.gz)

nix/crossBuild.nix
knuton: I would go with dynamic linking for Windows, it's worked well so far.

VERSION
knuton: Agree to move the version's source of truth to the flake file. I find it awkward to duplicate the version in the source and the tag, but we do that elsewhere and it works.

I think crossbuilding is not a release-specific activity. When I want to test something I may compile on my Linux machine and test on a Windows machine. I may not want to bother to set up the Windows machine to compile there. I would rather not have to commit each time I trigger a build.

There seems to be `dirtyRev` and `dirtyShortRev` too[^1][^2].

[^1]: https://discourse.nixos.org/t/flakes-accessing-selfs-revision/11237/7
[^2]: https://github.com/NixOS/nix/pull/5385

knuton: Oh, but we don't want `"2.3.0-${rev}"` when the rev is tagged, then we just want `"2.3.0"`.

I am not sure this is possible with flakes, because they make a point of throwing away useful information. :upside_down_face: 

How about

- making `flake.nix` the source of truth for the version,
- (while duplicating the version in a Git tag),
- only put a human-picked version string in the linked variable,
- include the revision as a separate variable and put into metadata [here](https://github.com/dividat/driver/blob/2aaf5787864a199a928404f3bea8cc659de52022/src/dividat-driver/server/main.go#L76-L82)

?

Makefile
knuton: Discussed here: https://github.com/dividat/driver/pull/129#discussion_r1554962503

Makefile
knuton: Probably irrelevant, see discussion here: https://github.com/dividat/driver/pull/129#discussion_r1554962503

VERSION
krksgbr: I see what you mean, I agree that keeping the `rev` as part of the version string is useful. Your suggestion sounded good but as I started working on it, I realized that it might not work so well compared to `git describe`. 
Checking out some tag, `git describe` give us just the tag, without anything extra:

```
❯ git checkout 2.3.0
Note: switching to '2.3.0'. 
❯ git describe
2.3.0
```

We can't replicate this with the information provided by nix unfortunately. I find the `<version>-<hash>` option which we can do with nix much less useful, because it doesn't tell us where we are in relation to the closest tag. Building from the tag would look pretty much the same as building from any later commit, we'd have to look at `git log` to make sense of it. We can't have 

I had another go at providing the cross-compilation toolchains through shells instead of building actual derivations and this time I got it to work. I think we can go with that and keep using `git describe` for determining the version.

nix/crossBuild.nix
krksgbr: The windows build was also statically linked previously, if I'm not mistaken:
https://github.com/dividat/driver/blob/2aaf5787864a199a928404f3bea8cc659de52022/Makefile#L77

https://github.com/dividat/driver/blob/2aaf5787864a199a928404f3bea8cc659de52022/Makefile#L29-L43

I'm not sure if it's really necessary though.

flake.nix
krksgbr: [Clarify flake description](https://github.com/dividat/driver/pull/129/commits/c22989ce7bac1aa469bfebfe6859e6400c823d68)

VERSION
krksgbr: [Provide cross-building toolchain in nix shells](https://github.com/dividat/driver/pull/129/commits/f8907ae140d171ca4444585e862ff1557fe1a9be)

nix/devShell.nix
krksgbr: This was made redundant by [Provide cross-building toolchain in nix shells](https://github.com/dividat/driver/pull/129/commits/f8907ae140d171ca4444585e862ff1557fe1a9be)

nix/crossBuild.nix
krksgbr: This was made redundant by [Provide cross-building toolchain in nix shells](https://github.com/dividat/driver/pull/129/commits/f8907ae140d171ca4444585e862ff1557fe1a9be)

nix/crossBuild.nix
krksgbr: This was made redundant by [Provide cross-building toolchain in nix shells](https://github.com/dividat/driver/pull/129/commits/f8907ae140d171ca4444585e862ff1557fe1a9be)

nix/driverBuildScript.nix
```diff
@@ -0,0 +1,71 @@
...
+, staticBuild
+}:
+let
+  staticLinkingLdFlags =
+    if staticBuild then
+      "-linkmode external -extldflags '-static'"
+    else
+      "";
+in
+(pkgs.writeShellApplication {
```
knuton, line 22: Could this just be a shell script?

I might be missing something, but it looks like the parameters could just be passed as CLI flags or env variables? Do we need to write a script file with Nix at all?

Makefile
```diff
@@ -65,16 +57,21 @@ record:
 record-flex:
 	@go run src/dividat-driver/recorder/main.go ws://localhost:8382/flex
 
+.PHONY: bin
```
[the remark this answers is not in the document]
krksgbr: It's not needed at all, I don't remember why I added it:
[Remove unneded bin target](https://github.com/dividat/driver/pull/129/commits/d09b902feed29b14422fc6f12deb3d8831fc84ff)

nix/driverBuildScript.nix
krksgbr: You're right, it doesn't have to be a nix script:
[Use regular shell script instead of nix generated script](https://github.com/dividat/driver/pull/129/commits/5eabfa62a5f3f82fb8486accf2e28b3d27ca723c)

knuton:
Nicely done! :+1:

PR #133, opened 2024-04-20
Use the platform for Promise and fetch

knuton:
Replace outdated test dependencies with builtin functionality of Node runtime.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -228,9 +228,9 @@
       "integrity": "sha512-yIovAzMX49sF8Yl58fSCWJ5svSLuaibPxXQJFLmBObTuCr0Mf1KiPopGM9NiFjiYBCbfaa2Fh6breQ6ANVTI0A=="
     },
     "node_modules/get-func-name": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/get-func-name/-/get-func-name-2.0.0.tgz",
-      "integrity": "sha1-6td0q+5y4gQJQzoGY2YCPdaIekE=",
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/get-func-name/-/get-func-name-2.0.2.tgz",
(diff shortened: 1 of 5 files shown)


krksgbr:
When running `rm -rf node_modules && make test` a few times, sometimes some of the Senso tests time out.
This issue seems to be present on `main` as well, so I don' think it's introduced here.

We're also still getting some vulnerability warnings when running `npm`. I tried `npm audit fix --force` which updates mocha (and transitive dependencies):

```diff
diff --git a/package.json b/package.json
index a81dbf2..3ecc4d6 100644
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "dependencies": {
     "bonjour": "^3.5.0",
     "chai": "^4.1.1",
-    "mocha": "^5.2.0",
+    "mocha": "^10.4.0",
     "ws": "^3.1.0"
   },
   "devDependencies": {
```

The tests still run the same, including the flakeyness:
```
rm -rf node_modules && npm audit fix --force && make test
```

Seeing that the flakeyness is not introduced by these changes and the tests still run as before, should we update `mocha` as well to get rid of the remaining warnings?

knuton:
> Seeing that the flakeyness is not introduced by these changes and the tests still run as before, should we update `mocha` as well to get rid of the remaining warnings?

Good idea, I hadn't even tried to do the forced update! https://github.com/dividat/driver/pull/133/commits/d21767c851aefa73f1eee4a4c234ad9d9418c767

knuton:
I didn't manage to trigger any test failures for just `make test` (does removing the `node_modules/` matter for triggering it?) on my machine. But seeing how some of the test utils are implemented using just heuristic sleeps, I am not surprised there may be flakiness.

krksgbr:
> I didn't manage to trigger any test failures for just `make test` (does removing the `node_modules/` matter for triggering it?) on my machine. But seeing how some of the test utils are implemented using just heuristic sleeps, I am not surprised there may be flakiness.

Running `make test` around 10 times with and without removing `node_modules`, it seemed that removing the directory increased the likelihood of the tests timing out. However I think this was just a random pattern and had nothing to do with `node_modules`. I can't see why there should be a connection.

krksgbr (approved):
👌

PR #134, opened 2024-04-20
Protect WebSocket endpoints from untrusted origin requests

knuton:
We previously added support for adding Private Network Access and CORS headers only for a whitelist of permissible origins. This protects browser users running the driver locally against abuse from third-party websites, which could make HTTP requests to the driver at the loopback address.

This commit extends this protection to WebSocket endpoints, for which browsers do not currently perform pre-flight requests. We therefore explicitly deny all requests from unknown origins.

This change prepares for the driver gaining the capability of triggering firmware updates in connected devices (#127).

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,7 @@
 
 ### Changed
 
+- Deny WebSocket connections from unknown origins
 - Let firmware update command look for bootloader if no Senso in regular mode is found
 - Update build system and development environment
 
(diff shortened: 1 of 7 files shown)


krksgbr (approved):
It looks good and works as expected.

How I tested:
- I ran the tests with a build from `main` and the unknown origin tests failed, as expected.
- I also tested in the browser with the driver started with `./bin/dividat-driver --permissible-origin "http://orig.in"` and all requests were rejected, as expected.
- When the driver is started with no arguments, connections continue to work as before.

PR #135, opened 2024-04-23
Add direct dependency minimist

knuton:
Previously brought in as transitive dependency, but used as direct dependency of replay tools.

This makes the dependency explicit.

## Checklist

-   [x] Changelog updated
-   [x] Code documented


krksgbr (approved):
👌

PR #136, opened 2024-04-23
Add mode field in replayer

knuton:
All modern Sensos send this field and we are making the driver expect it.

@krksgbr I know you have something in the pipeline, but this seems like an easy immediate fix until then.

## Checklist

-   [x] Changelog updated
-   [x] Code documented


krksgbr (approved):
👌

PR #137, opened 2024-04-23
Fix potential index out of range error

krksgbr:
Being a go noob, I misunderstood how to safely check if an element is present in a slice.
Trying to access `slice[0]` when the slice is empty doesn't evaluate to `nil`, but causes an index out of range error.
We haven't noticed this because the condition never came about, but now it did while running some tests with the Senso mocking tool I've been working on.

## Checklist

-   ~[ ] Changelog updated~
-   ~[ ] Code documented~


knuton (approved):
:+1:

PR #139, opened 2024-05-22
Clarify message about channel head

knuton:
Add "currently" to avoid confusion between what is on the channel and what is about to be deployed.


krksgbr (approved):
👍

PR #142, opened 2024-08-27
Increase discovery timeout to 120 s in firmware updater

### LINKED ISSUE #141: Extend discoveryTimeout in firmware update routine
The [discovery timeout is currently 60 s](https://github.com/dividat/driver/blob/22faad9999ddd88222b9f010289f942969c68910/src/dividat-driver/firmware/main.go#L33), which is seemingly occasionally too low when rediscovering the Senso in bootloader mode.

On Aug 22 I observed update failure for three Sensos, all connected via direct Ethernet link to a Play Computer running PlayOS 2024.7.0. The update completed upon retry, presumably the Sensos had finished booting and announcing their service by then.

knuton:
There are at least three instances (albeit all at the same customer location) of firmware update initially failing due to discovery timeout while waiting for the bootloader to appear. In all cases, the update completed when retried a brief while later, as the bootloader had finished startup and announcing itself via mDNS, plus that being picked up by the client.

Based on the timestamps of my photos from the site, there must have been around 20 seconds between the initial failure and eventual bootloader discoverability. So 90 seconds timeout may be enough, but it is not worth trading unnecessary update failures for occasional 30 seconds of unsuccessful waiting. Giving two full minutes for discovery.

Fixes #141.

## Checklist

-   [x] Changelog updated
-   [x] Code documented

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 ## [UNRELEASED]
 
+### Changed
+
+- Allow up to 2 minutes for device discovery in firmware updater
+
 ## [2.4.0] - 2024-05-22
 
 ### Added
--- a/src/dividat-driver/firmware/main.go
+++ b/src/dividat-driver/firmware/main.go
@@ -30,7 +30,7 @@ import (
 
 const tftpPort = "69"
 const controllerPort = "55567"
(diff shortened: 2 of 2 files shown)


krksgbr (approved):
I can't think of anything that would speak against this change.

PR #146, opened 2025-03-21
Add bitdepth command and bitdepth=12 support to Flex driver

yfyf:
See commit messages for more details. They are to be squashed, but keeping them separate so you can see the initial "naive" implementation I did and then the horrendous hack I had to resort to.

If you want to try it with Play, use the [flex-v5 branch](https://github.com/yfyf/diviapps/tree/flex-v5)

### DIFF
--- a/src/dividat-driver/flex/main.go
+++ b/src/dividat-driver/flex/main.go
@@ -15,6 +15,7 @@ The functionality of this module is as follows:
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/binary"
 	"strings"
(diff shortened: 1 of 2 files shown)


src/dividat-driver/flex/main.go
```diff
@@ -161,98 +193,175 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
...
-			case i := <-tx:
-				data, _ := i.([]byte)
+					logger.WithFields(logrus.Fields{
+						"old": configuredBytesPerSample,
+						"new": newBytesPerSample,
+					}).Info("Bytes per sample changed, will restart reader and controller")
+
+					logger.Info("Closing port and waiting for reader to stop")
+					port.Close()
+					<-readerDoneChan
```
yfyf, line 253: I'm not sure this is strictly necessary - once we close the port, the reader dies and that's it. But I think it's a good to get an ack here, so that we don't accidentally spawn multiple readers or anything like that, in case of future refactoring and whatnot?

src/dividat-driver/flex/main.go
```diff
@@ -161,98 +193,175 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
...
 	for {
 		// Terminate if we were cancelled
 		if ctx.Err() != nil {
+			logger.Debug("Stopping reader: context cancelled")
 			return
 		}
 
 		input, err := reader.ReadByte()
 		if err != nil {
+			logger.WithField("err", err).Error("Error reading from serial port")
```
yfyf, line 338: When the port is closed on bitdepth change, this will produce a a single error log. I think this is preferable to silently dying though, because there might be other situations when reading fails that we might want to know about.

PR #148, opened 2025-05-16
Add configurable bitdepth support to the Flex driver

yfyf:
Necessary refactoring: swap the input stream reading and client command forwarding coroutine nesting in order to be able to reset the input stream upon bitdepth change.

Restarting the reader is a both a clean approach to reseting the protocol parsing state machine and also solves a particular issue of the reader hanging on insufficient input in the following scenario:

- Driver is running at 8bit depth
- Reader is midway reading a frame
- Driver/Firmware is updated to run at 12bit depth
- Reader now expects the next sample to be 12 bits, but the current buffered frame still contains an 8bit sample.
- Reader is now stuck on `reader.ReadByte()` forever

These changes are intended to be backwards compatible with older versions of Play (which send no bitdepth commands and assume that the driver sets 8bit mode by default).

## TODOs

- [ ] Test with Flex v4
- [ ] Test with Flex v5
- [ ] Test with old Play
- [ ] Test with new Play
- [ ] Ask @dividat-jgu to test on his Flex

## Checklist

- [ ] Changelog updated
- [ ] Code documented


src/dividat-driver/flex/main.go
```diff
@@ -161,98 +190,159 @@ func connectSerial(ctx context.Context, logger *logrus.Entry, serialName string,
...
-	// We hardcode a bitdepth of 8 for sample acquisition.
-	// In principle this could be made configurable and left to the client.
-	// However, parsing of the byte stream requires knowing the bitdepth,
-	// so in order to assemble frame packages the driver would need to
-	// intercept client-to-device commands and configure the parser
-	// accordingly. As we don't need acquisition at other than 8 bits it
-	// seems more robust to fix the mode in the driver right now.
-	BYTES_PER_SAMPLE := 3 // Row, column and sample value of 8 bit
-	BITDEPTH_8_CMD := []byte{'U', 'L', '\n'}
+	// For backwards compatibility, we set 8bit depth by default
```
yfyf, line 208: Unsure about this. Alternatives:
- Set bitdepth=12 by default to match the Flex v5 default
- Do not reconfigure the bitdepth by default at all

I am thinking that maybe the latter is best?

knuton: Without having thought through the implications fully, the latter indeed sounds best in the current landscape, makes the least assumptions.

yfyf:
> Looks good and nicely structured from a first look, wonder what you think of the possibility of deferring serial reader setup until we get profile from client. I think the biggest Achilles' heel of that idea would be that we don't have any guarantee that client will send a bitdepth command, and then we'd still have to pick a default?

I like the idea of only sending data after proper setup is done, but I think several arguments against it suggest we don't want to do that:
- It would break simple debugging (plain WS connections) and tools like `record-flex`
- "until we get profile" is not clearly defined - is it just the bitdepth? what about the other commands?

This would introduce an extra "init protocol" that further adds to the mess/complexity of interacting with the current firmware/hardware, so I would rather just not do it. It implies both driver and client have to correctly signal and respond to connection resets.

src/dividat-driver/flex/main.go
yfyf: Hmmm, I was about to remove the default bitdepth, but I realized that's not an option unless we also don't start the reader by default, which, as discussed earlier, leads to other complications.

I suggest we let's leave it as is, this setup already received significant testing and arguments for changing it are kinda weak, or set 12bit bitdepth as default (assuming that the new driver will never be used with older versions of Play).

knuton: OK

knuton:
I confirmed that the reader routine was restarted appropriately according to the bitdepth sent using a Teensy flashed with v5 firmware.

PR #174, opened 2026-05-22
Avoid doing git operations after checkout

yfyf:
Noticed that this breaks builds in private repos, since without credentials it is not possible to fetch the tags.

This is future-proofing the usage of `persist-credentials: false`. For a public repo, this changes nothing. But if `driver` were to become a private repo or something in Github's logic were to change to block public reads, this would start failing.


yfyf:
Release tag run on my fork: https://github.com/yfyf/driver/actions/runs/26278423003

knuton (approved):
Good, thanks

PR #179, opened 2026-06-25
Update Flex V6 recordings using a real mat

yfyf:
Previous recordings were made with a prototype kit.

Dropping the `.ws` recordings, since they are redundant in the case of V6 devices.

## Checklist

-   ~~Changelog updated~~
-  ~~Code documented~~


knuton (approved):
👍 

Checked replayed recordings via Play settings

